PR #2, opened 2017-01-09
Refactor of tree mesh

rowanc1:


lheagy:
@rowanc1: any concerns with deleting this branch?

PR #3, opened 2017-01-09
Initial work on the octree creation example.

rowanc1:


lheagy:
@rowanc1: Is this still on the radar? or should we close and abandon?

lheagy:
I am going to close this as there hasn't been activity in quite a while. Feel free to re-open if you would like to revive it

PR #4, opened 2017-01-09
Pypi Release 0.1.0

rowanc1:


rowanc1:
@lheagy can you take a quick look at this before I merge to master.

`pip install discretize` should work now.

You may need an updated pymatsolver. (also works on pip)

lheagy:
It doesn't look like the pip install is working (microsoft azure notebooks - runs linux) 

image

Makefile
```diff
@@ -0,0 +1,32 @@
...
+clean:
+	cd docs;make clean
+	find . -name "*.pyc" | xargs -I {} rm -v "{}"
+
+clean_cython: clean
+	find . -name "*.c" | xargs -I {} rm -v "{}"
+	find . -name "*.so" | xargs -I {} rm -v "{}"
+
+deploy:
+	python setup.py sdist bdist_wheel upload
```
lheagy: do we want to use `twine` ?
https://packaging.python.org/distributing/

image

discretize/utils/interputils.py
```diff
@@ -66,99 +66,3 @@ def interpmat(locs, x, y=None, z=None):
     Q = sp.csr_matrix((vals, (I, J)),
                       shape=(npts, np.prod(shape)))
     return Q
-
-if not _interpCython:
```
lheagy: Until we get the pip install stable with cython, I think we should hang on to this code...

docs/content/installing.rst
```diff
@@ -19,6 +19,7 @@ Development Dependencies
 - sphinx
 - sphinx_rtd_theme
 - sphinx-gallery
```
lheagy: should we pull these into a dev-requirements.txt (sphinx and sphinx-gallery are not strictly necessary for discretize)

discretize/utils/meshutils.py
```diff
@@ -181,11 +181,11 @@ def ExtractCoreMesh(xyzlim, mesh, meshType='tensor'):
     This function ouputs::
 
         - actind: corresponding boolean index from global to core
-        - meshcore: core SimPEG mesh
+        - meshcore: core mesh
 
     Warning: 1D and 2D has not been tested
     """
-    from SimPEG import Mesh
+    from discretize import Mesh
```
lheagy: this will fail

docs/content/installing.rst
rowanc1: they are in a req_dev.txt already.

discretize/utils/meshutils.py
rowanc1: fixing this. (it is not tested....)

docs/content/installing.rst
lheagy: kk, do we want them in the `installing.rst`?

rowanc1:
@lheagy I got the build semi working on appveyor. Looks like the cython is installing at least.

There is a problem in the TreeMesh with integer overflows, otherwise things seem to work. I am not sure if the `pip` install will work on windows, but it does work on linux now.

I am going to leave the int overflow (I think it needs to be 64bit?) for another time.

.bumpversion.cfg
```diff
@@ -1,4 +1,4 @@
 [bumpversion]
-current_version = 0.0.5
+current_version = 0.1.1b1
```
lheagy: does this do the `--pre`?

rowanc1: yeah

rowanc1:
image

PR #9, opened 2017-01-10
Windows/dev

jcapriot:
Changes for windows compatibility:
 - `long` changed to `long long`

Changes for appveyor testing:
- needed to cd into a different directory to do the tests

Setup.py changes:
- Used numpy's distutils for package management

This can now be built and installed without any special arguments to the setup.py file. For example, python setup.py build and python setup.py install
both work.

The setup.py files within each subpackage take care of cythonizing and attaching the extensions, making the main setup.py file a little cleaner.

This version was able to be pip installed on my Windows machine, should consider building wheels for distribution without compilation.


jcapriot:
This should make the main simpeg branch pure python!

rowanc1:
@jcapriot thanks for your help on this!

PR #11, opened 2017-01-16
Feat/view

lheagy:
add down-sampling capability to _plot2DImage as per simpeg/simpeg#537


rowanc1:
@lheagy is this what will fix this? If so, maybe we can move the example from simpeg to here as well on this PR?

See the SimPEG PR: simpeg/simpeg#542

image

lheagy:
@rowanc1 : thoughts on bringing this in?

Issue #23, opened 2017-01-17
Interpolation


rowanc1:
_From @rowanc1 on November 17, 2015 22:0_

In the interp utils we should be using `np.searchsorted` rather than `np.argmin` over the vectors:

Speed up of O(log(n)) vs O(n) for each point you are interpolating, could be significant, especially on large meshes.

_Copied from original issue: simpeg/simpeg#170_

jcapriot:
I think we could actually get away without having the interputil_cython extension if you used `np.searchsorted` as that function accepts a list of points to insert.

PR #28, opened 2017-02-06
implementing bisect algorithm for searching for point indexes without…

jcapriot:
implementing the bisection algorithm without the GIL. Gets rid of a lot of python overhead for speed-ups. If using Cython, might as well use it to get down to C-like code.

All of the `for` loops are executed without any calls to the python interpreter, and the bisection algorithm is equivalent to the np.searchsorted without the python overhead. Also, since we know how big the `inds` and `vals` array will be, might as well pre-allocate them.

At least 2x-3x speed on decent sized arrays.

Should address #23


jcapriot:
i'll fix this tonight... not quite sure why it broke tree tests though.

rowanc1:
This looks awesome. I don't have a lot of experience with cython. Any chance you can give a brief overview tomorrow?

jcapriot:
TreeMesh could use a lot of love to optimize it up.

lheagy:
This looks solid @jcapriot! For an overview see https://youtu.be/zlnfQMzEOEA

lheagy:
@rowanc1, @jcapriot: are you good with merging this??

rowanc1:
@jcapriot can you comment?

The cython one that I made, is all yellow (python), @jcapriot wrote it so that it actually used C. 👍 

image

image

jcapriot:
I'm good with it.
Mostly I just got around the python calls while inside the loops to make them faster!

PR #35, opened 2017-02-27
give the Identity a transpose

lheagy:
- give the `Identity` class a `T` and `transpose`
- return `self` when transposing a `Zero`

### DIFF
--- a/discretize/utils/matutils.py
+++ b/discretize/utils/matutils.py
@@ -478,10 +478,12 @@ class Zero(object):
         return 0 > v
 
     @property
-    def transpose(self): return Zero()
+    def transpose(self):
+        return self
 
     @property
-    def T(self): return Zero()
+    def T(self):
+        return self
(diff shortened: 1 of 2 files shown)


discretize/utils/matutils.py
```diff
@@ -561,3 +563,11 @@ def __ge__(self, v):
 
     def __gt__(self, v):
         return 1 > v if self._positive else -1 > v
+
+    @property
+    def T(self):
+        return self
+
+    @property
+    def transpose(self):
```
rowanc1: I think that transpose should be a function not a property:

https://github.com/scipy/scipy/blob/v0.14.0/scipy/sparse/csr.py#L108

lheagy: kk, sounds good

lheagy:
@rowanc1 are you good with these changes?

PR #36, opened 2017-03-03
3D Cyl Mesh

lheagy:
# 3D Cyl Mesh
First implementation of a 3D cylindrical mesh. We discretize in r, theta and z.
image

## Operators
- `edgeCurl`
- `faceDiv`
- `aveF2CC`
- `aveF2CCV`
- `aveE2CC`
- `aveE2CCV`

are all implemented and tested for second order convergence

## Examples
There are some examples running this code in http://github.com/lheagy/casingResearch
image
- added a simple example that plots a 3D cyl mesh

## Plotting
- add plot capabilities for plotSlice (and I think two slices through the mesh for `plotGrid()`)

## Testing
- basic elements like the grids, counting of things. 
- operator tests for mimetic properties

# Other updates

## Cell Grad Stencils are properties
- cell grad stencils are properties rather than methods. This pr will need to be followed by an update in the SimPEG regularization (addressed in simpeg/simpeg#699)

## minor updates
- plotting now uses the `C0`, `C1`, ... colors in matplotlib
- added `*.ipynb` to .gitignore


rowanc1:
I would like to keep SimPEG out of discretize. That being said, the examples can point over to that repository/documentation. Similar to pointing out to current other research that is ongoing.

discretize/View.py
```diff
@@ -646,7 +648,120 @@ def _plotCylTensorMesh(self, plotType, *args, **kwargs):
...
+            return self._plotCylTensorMesh('plotGrid', *args, **kwargs)
+
+        # allow a slice to be provided for the mesh
+        slc = kwargs.pop('slice', None)
+        if isinstance(slc, str):
+            slc = slc.lower()
+        assert slc in ['theta', 'z', 'both', None], (
+            "slice must be either 'theta','z', or 'both' not {}".format(slc)
+        )
+
```
lheagy: image

rowanc1: What does `both` do?

lheagy: same thing as `mesh.plotGrid()` plots both the `theta` and `z`

discretize/View.py
```diff
@@ -646,7 +648,120 @@ def _plotCylTensorMesh(self, plotType, *args, **kwargs):
...
+                    polarax = [a for a in ax if a.__class__.__name__ == 'PolarAxesSubplot']
+                    if len(polarax) != 1:
+                        warnings.warn("""
+No polar axes provided. Over-writing the axes. If you prefer to create your
+own, please use
+
+    `ax = plt.subplot(121, projection='polar')`
+
+for reference, see: http://matplotlib.org/examples/pylab_examples/polar_demo.html
+                    https://github.com/matplotlib/matplotlib/issues/312
```
lheagy: it is a little annoying, the api for polar plots is more restricted. I have not found a way to do
```
fig, ax = plt.subplots(1, 2, figsize=(12, 4))
ax[0].make_this_polar
```

lheagy: see matplotlib/matplotlib#312

requirements_dev.txt
```diff
@@ -14,3 +14,4 @@ ipython
 matplotlib
 properties[math]
 wheel
+pytest
```
lheagy: @rowanc1: thoughts on bringing this in as a dev dependency? It makes testing that warnings are thrown easier. http://doc.pytest.org/en/latest/

tests/base/test_cyl3D.py
```diff
@@ -0,0 +1,702 @@
...
+            np.all(
+                self.mesh.vol ==
+                np.pi * self.mesh.hx ** 2 * self.mesh.hz / self.mesh.nCy
+            )
+        )
+
+# ----------------------- Test Grids and Counting --------------------------- #
+
+
+class Cyl3DGrid(unittest.TestCase):
```
lheagy: I would appreciate another set of eyes here. I tried to pick a simple example that demonstrated how things are counted. See what you think

tests/base/test_cyl.py
```diff
@@ -586,7 +596,7 @@ def test_nC(self):
         self.assertTrue(np.all(self.mesh.vnC == [3, 2, 2]))
 
     def test_nN(self):
-        self.assertTrue(self.mesh.nN == 24)
+        self.assertTrue(self.mesh.nN == 21)
```
lheagy: this is a change in assumption on the counting in the 3D cyl mesh

tests/base/test_cyl.py
```diff
@@ -608,22 +618,22 @@ def test_nE(self):
         self.assertTrue(self.mesh.nEy == 18)
         self.assertTrue(np.all(self.mesh.vnEy == [3, 2, 3]))
         self.assertTrue(self.mesh.nEz == 12 + 2)
-        self.assertTrue(self.mesh.vnEz is None)
+        self.assertTrue(np.all(self.mesh.vnEz == [4, 2, 2]))
```
lheagy: this is also a change in assumption on counting

rowanc1: This is only for the 3D one though correct?

For the symmetric one, there should still be no vertical edges.

lheagy: correct. This is testing the 3D. There are some tests in `test_cyl` that test the construction of a 3D cyl mesh (I didn't notice them before)

lheagy:
The current implementation of the interpolation to a cartesian mesh is limited to cylindrically symmetric cases. The extension to the full 3D is non-trivial, and not strictly necessary for running simulations. I would vote for addressing it in a separate pr.

lheagy:
Looks like Appveyor is running out of memory. Perhaps we should choose a smaller subset of tests to run there? 
image

.travis.yml
```diff
@@ -29,9 +29,10 @@ install:
   - pip install -r requirements_dev.txt
   - python setup.py install
   - make build
+  - export MPLBACKEND="agg"
 
 script:
-  - nosetests tests/base tests/tree tests/docs --with-cov --cov discretize --cov-config .coveragerc -v -s;
+  - nosetests tests/base tests/tree tests/docs tests/cyl --with-cov --cov discretize --cov-config .coveragerc -v -s;
```
lheagy: I parsed out the cyl tests so that they don't run on appveyor - they seem to cause memory issues (seems to be related to the import of matplotlib?) Travis is happy though :)

discretize/InnerProducts.py
```diff
@@ -12,7 +12,11 @@ class InnerProducts(object):
     This mixIn creates the all the inner product matrices that you need!
     """
     def __init__(self):
-        raise Exception('InnerProducts is a base class providing inner product matrices for meshes and cannot run on its own. Inherit to your favorite Mesh class.')
+        raise Exception(
+            'InnerProducts is a base class providing inner product matrices '
```
lheagy: just whitespace cleanup

lheagy:
@rowanc1: If you have a chance to review this, I would appreciate hearing your thoughts!

lheagy:
I would also like to add a cell grad on this mesh. I am not sure if that should be included in this pr, what do you think @rowanc1?

lheagy:
I am relatively happy with this as a first pass feature addition - the old functionality of the cyl symmetric mesh is there, and now the 3D cyl divergence and curl are tested and working as I would expect. @rowanc1, can you take a look when you have a chance and let me know if you see any glaring oversights? There is certainly more to be done, but I think further changes get outside the scope of this pr.

discretize/CurvilinearMesh.py
```diff
@@ -304,9 +304,18 @@ def normals(self):
         if self.dim == 2:
             return normalize2D(np.r_[self._normals[0], self._normals[1]])
         elif self.dim == 3:
-            normal1 = (self._normals[0][0] + self._normals[0][1] + self._normals[0][2] + self._normals[0][3])/4
-            normal2 = (self._normals[1][0] + self._normals[1][1] + self._normals[1][2] + self._normals[1][3])/4
-            normal3 = (self._normals[2][0] + self._normals[2][1] + self._normals[2][2] + self._normals[2][3])/4
+            normal1 = (
```
lheagy: this is just whitespace cleanup

discretize/CylMesh.py
```diff
@@ -26,36 +30,50 @@ class CylMesh(BaseTensorMesh, BaseRectangularMesh, InnerProducts, CylView):
...
+        )
         self.cartesianOrigin = np.array(cartesianOrigin, dtype=float)
 
     @property
     def isSymmetric(self):
         return self.nCy == 1
 
     @property
+    def _ntNx(self):
+        if self.isSymmetric:
```
lheagy: eventually, I would like to remove this assumption (there is conceptually a node at the origin, even for the cyl symmetric case, we just remove it). For the cyl symmetric code though, this assumption is used pretty heavily at the moment, so I think it warrants a separate pr

discretize/CylMesh.py
```diff
@@ -123,72 +227,531 @@ def vectorCCx(self):
...
+                    )
+                )
         return self._vol
 
     ####################################################
+    # Active and Hanging Edges and Faces
+    ####################################################
+
+    @property
+    def _ishangingFx(self):
```
lheagy: I am not sure if krons of bools is the most elegant way to do this, but it was the clearest in my head. I would appreciate hearing thoughts!

discretize/CylMesh.py
```diff
@@ -200,48 +763,96 @@ def faceDiv(self):
...
+
+    #     if self.isSymmetric:
+    #         G1 = sp.kron(utils.speye(n[2]), ddxCellGrad(n[0], BC))
+    #     else:
+    #         G1 = self._deflationMatrix('Fx').T * utils.kron3(
+    #             utils.speye(n[2]), utils.speye(n[1]), ddxCellGrad(n[0], BC)
+    #         )
+    #     return G1
+
+    @property
```
lheagy: - [ ] TODO: this should raise a `NotImplementedError` right now - it is not tested

discretize/CylMesh.py
```diff
@@ -263,111 +874,234 @@ def nodalLaplacian(self):
...
-            Dr = sp.kron(sp.identity(self.nNz), dr)
-            Dz = -sp.kron(dz, sp.identity(self.nCx))
-
             A = self.area
             E = self.edge
-            # Edge curl operator
-            self._edgeCurl = (utils.sdiag(1/A)*sp.vstack((Dz, Dr)) *
-                              utils.sdiag(E))
+
+            if self.isSymmetric is True:
```
lheagy: again, I would like to use deflation matrices for this, but it is a bit out of scope for this pr

requirements_dev.txt
rowanc1: That sounds great!

discretize/CylMesh.py
rowanc1: `krons of bools`, how very cryptic of you. I like how this has become the easy way to do things...

rowanc1:
I have taken a brief look. I will look more in depth tomorrow.

lheagy:
I think the cyl to cart interpolation is out-of-scope here at the moment. That can be addressed in a separate pr later on

discretize/DiffOperators.py
```diff
@@ -302,21 +447,57 @@ def setCellGradBC(self, BC):
...
+    @property
+    def _cellGradzStencil(self):
+        if self.dim < 3:
+            return None
+        BC = ['neumann', 'neumann'] # TODO: remove this hard-coding
+        n = self.vnC
+        G3 = kron3(ddxCellGrad(n[2], BC), speye(n[1]), speye(n[0]))
+        return G3
+
+    @property
```
lheagy: I made this a property. Previously it was a function with no inputs, but it makes more sense as a property. This will require an update in the regularization mesh over in SimPEG

lheagy:
@rowanc1: would you mind taking a look at this? I have commented on the things that I think are most important to have a review of

discretize/CylMesh.py
rowanc1: What does the `nt` stand for?

discretize/CylMesh.py
```diff
@@ -123,72 +227,531 @@ def vectorCCx(self):
...
+            hang_x = np.zeros(self._ntNx, dtype=bool)
+            hang_x[0] = True
+            self.__ishangingFx = np.kron(
+                np.ones(self.nCz, dtype=bool),
+                np.kron(
+                    np.ones(self.nCy, dtype=bool),
+                    hang_x
+                )
+            )
+        return self.__ishangingFx
```
rowanc1: Double underscores are not ideal to use. Not an issue here, but good to stay aware of.


> Private name mangling: When an identifier that textually occurs in a class definition begins with two or more underscore characters and does not end in two or more underscores, it is considered a private name of that class. Private names are transformed to a longer form before code is generated for them. The transformation inserts the class name in front of the name, with leading underscores removed, and a single underscore inserted in front of the class name. For example, the identifier __spam occurring in a class named Ham will be transformed to _Ham__spam. This transformation is independent of the syntactical context in which the identifier is used. If the transformed name is extremely long (longer than 255 characters), implementation defined truncation may happen. If the class name consists only of underscores, no transformation is done.

https://docs.python.org/3.5/reference/expressions.html?highlight=mangling#index-5

discretize/CylMesh.py
```diff
@@ -123,72 +227,531 @@ def vectorCCx(self):
...
+            ])
+            deflate = zip(
+                np.nonzero(self._ishangingEz)[0].tolist(), deflateEz
+            )
+
+            self.__hangingEz = dict(deflate)
+        return self.__hangingEz
+
+    @property
+    def _axis_of_symmetry_N(self):
```
rowanc1: I am not entirely sure what this is for. Can you add a doc string?

discretize/CylMesh.py
rowanc1: Does the `_cellGradxStencil` exist? in which case will this fail anyways?

discretize/TensorMesh.py
```diff
@@ -594,93 +598,199 @@ def vol(self):
...
+    @property
     def area(self):
         """Construct face areas of the 3D model as 1d array."""
-        if getattr(self, '_area', None) is None:
+        if self.dim == 1:
+            return self.areaFx
+        elif self.dim == 2:
+            return np.r_[self.areaFx, self.areaFy]
+        elif self.dim == 3:
+            return np.r_[self.areaFx, self.areaFy, self.areaFz]
```
rowanc1: Does `r_` have the same performance hit that `hstack` did? Likely not?

discretize/View.py
```diff
@@ -581,14 +593,15 @@ def _plotCylTensorMesh(self, plotType, *args, **kwargs):
         # Just create a TM and use its view.
         from discretize import TensorMesh
 
-        vType = kwargs.pop('vType', None)
+        vType = kwargs.get('vType', None)
```
rowanc1: Is there a reason this got changed to `get` from `pop`?

discretize/TensorMesh.py
lheagy: both seem to be fine for vectors
image

discretize/CylMesh.py
lheagy: ah, kk, thanks. I can update that

discretize/CylMesh.py
lheagy: ha, is that horrible??

discretize/CylMesh.py
lheagy: - [ ] todo: docstring here. This identifies the nodes along the axis of symmetry that need to be eliminated

discretize/View.py
lheagy: yeah, for 'CC', we don't reassign vType, and it still needs to be passed to the tensor plot

lheagy:
I would like to get #72 in before we merge this so that the changes wrt having `cellGradStencil` as a property can be done on this pr (this will also require an update in SimPEG regularization)

lheagy:
Also on this pr, I have updated the colors for `mesh.plotGrid()` so that the more recent matplotlib color palette is being used

image

Please let me know if this is ok with you all :)

.travis.yml
```diff
@@ -2,17 +2,22 @@ language: python
...
-    - python: 3.7
+    - python: "nightly"
 
 sudo: false
 
 env:
   global:
     - MASTER_BRANCH=master
     - PYPI_PY=3.6  # deploy to pypi from python 3.6
+  matrix:
```
lheagy: I split up the tests here (it is helpful for diagnosing failures to have them be tested separately)

lheagy:
I would like to get this in soon, and simpeg/simpeg#683 to follow in SimPEG. @rowanc1, would you mind taking another pass through this one?

rowanc1:
This is pretty big now. I can try to play around with it tonight or tomorrow.

lheagy:
It is, thanks @rowanc1. Feel free to give me a shout if you would like to chat about it!

lheagy:
@rowanc1: would you have time to look at this?

discretize/CylMesh.py
```diff
@@ -146,72 +328,715 @@ def vectorCCx(self):
...
+
+        :rtype: numpy.ndarray
+        :return: grid locations of radial faces
+        """
+        if getattr(self, '_gridEz', None) is None:
+            if self.isSymmetric is True:
+                self._gridEz = None
+            else:
+                self._gridEz = (
+                    self._deflationMatrix('Ez', withHanging=False) *
```
rowanc1: This is just a delete the rows operation?

discretize/CylMesh.py
```diff
@@ -285,112 +1189,294 @@ def nodalLaplacian(self):
...
+                    ])
+            return sp.block_diag([
+                self._deflationMatrix(
+                    location+coord, withHanging=withHanging, asOnes=asOnes
+                )
+                for coord in ['x', 'y', 'z']
+            ])
+
+        R = speye(getattr(self, '_nt{}'.format(location)))
+        hanging = getattr(self, '_hanging{}'.format(location))
```
rowanc1: `hanging_dict`? might be clearer

discretize/CylMesh.py
```diff
@@ -285,112 +1189,294 @@ def nodalLaplacian(self):
...
+        nothanging = ~getattr(self, '_ishanging{}'.format(location))
```
rowanc1: Might be better to use this sort of logic on the grids. Rather than construct the deflated identity matrix.

discretize/CylMesh.py
```diff
@@ -285,112 +1189,294 @@ def nodalLaplacian(self):
...
+                    (self.aveFx2CC, self.aveFy2CC, self.aveFz2CC),
+                    format="csr"
+                )
         return self._aveF2CCV
 
+    ####################################################
+    # Deflation Matrices
+    ####################################################
+
+    def _deflationMatrix(self, location, withHanging=True, asOnes=False):
```
rowanc1: Remove `withHanging` option?

discretize/CylMesh.py
```diff
@@ -146,72 +328,715 @@ def vectorCCx(self):
...
+            if self.isSymmetric:
+                self._ishangingEzBool = np.ones(self._ntEz, dtype=bool)
+            else:
+                hang_x = np.zeros(self._ntNx, dtype=bool)
+                hang_x[0] = True
+
+                hang_y = np.zeros(self._ntNy, dtype=bool)
+                hang_y[-1] = True
+
+                hangingEz = np.kron(
```
rowanc1: I think this is exactly the same?
```
hanging = np.zeros(self._vntEz, dtype=bool)
hanging[0, :, :] = True  # The first z-edges are the same
hanging[:, -1, :] = True  # All the z-edges wrap
hanging[0, 0, :] = False  # Not the axis of symmetry
return mkvc(hanging)
```

discretize/CylMesh.py
```diff
@@ -146,72 +328,715 @@ def vectorCCx(self):
...
+            )
+        return self._hangingFyDict
+
+    @property
+    def _ishangingFz(self):
+        """
+        bool vector indicating if a z-face is hanging or not
+        """
+        if getattr(self, '_ishangingFzBool', None) is None:
+            self._ishangingFzBool = np.kron(
```
rowanc1: can probably just do a `np.zeros` here instead of doing krons.

discretize/CylMesh.py
lheagy: yep, good call, that is much simpler

discretize/CylMesh.py
lheagy: 👍

discretize/CylMesh.py
lheagy: It is, but unfortunately working in 3D and then flattening is way less efficient 
image

discretize/CylMesh.py
```diff
@@ -223,58 +1048,137 @@ def faceDiv(self):
...
-            self._faceDivz = utils.sdiag(1/V)*D3*utils.sdiag(S)
+            self._faceDivz = sdiag(1/V)*D3*sdiag(S)
         return self._faceDivz
 
+    # @property
+    # def _cellGradxStencil(self):
+    #     n = self.vnC
+
+    #     if self.isSymmetric:
+    #         G1 = sp.kron(speye(n[2]), ddxCellGrad(n[0], BC))
```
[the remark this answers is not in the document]
lheagy: I haven't tested it yet. I think that cellGrads are beyond the scope of this pr. They are only used in regularization, not in the forward simulations

lheagy:
@rowanc1: I simplified a few things: using indices instead of the deflation matrix on the grids, edges and areas. I think we need to stick with the krons of bools though rather than the 3D matrix implementation - there is a big efficiency difference, and this really matters for the larger casing meshes

discretize/utils/coordutils.py
```diff
@@ -2,18 +2,74 @@
...
+        mkvc(points[:, 2], 2)
+    ])
 
-        :param numpy.array v0: vector of length 3
-        :param numpy.array v1: vector of length 3
-        :param tol = 1e-20: tolerance. If the norm of the cross product between the two vectors is below this, no rotation is performed
-        :rtype: numpy.array, 3x3
-        :return: rotation matrix which rotates the frame so that n0 is aligned with n1
 
+def rotate_vec_cyl2cart(grid, vec):
```
lheagy: This should eventually be moved to `matrixutils`

lheagy:
For versioning, I think this should be a `coda minor` e.g. 
```discretize 0.1.19 --> 0.2.0```
as it contains non-backwards compatible changes with the cell grad stencil usage and is a fairly major addition

Thoughts??

File: discretize/CylMesh.py [hunk 28]
```diff
@@ -434,10 +421,7 @@ def edgeEz(self):
         """
         if getattr(self, '_edgeEz', None) is None:
 
-            self._edgeEz = (
-                self._deflationMatrix('Ez', withHanging=False) *
-                self._edgeEzFull
-            )
+            self._edgeEz = self._edgeEzFull[~self._ishangingEz]
```
rowanc1: These are much cleaner!

discretize/CylMesh.py
```diff
@@ -614,6 +588,12 @@ def vol(self):
 
     ####################################################
     # Active and Hanging Edges and Faces
+    #
+    #    - to find the active edges, faces, we use
+    #      krons of bools (sorry). It is more efficient
+    #      than working with 3D matrices. For an
+    #      example, see the comments in `_ishangingFx`
```
rowanc1: Can we have the code example in here showing the equivalency? The note from the PR discussion between you and I should do well.

discretize/CylMesh.py
```diff
@@ -622,10 +602,18 @@ def _ishangingFx(self):
         bool vector indicating if an x-face is hanging or not
         """
         if getattr(self, '_ishangingFxBool', None) is None:
+
+            # the following is equivalent to
+            #     hang_x = np.zeros(self._vntFx, dtype=bool)
```
rowanc1: Totally cool. Looks like you are already adding these!!

discretize/CylMesh.py
```diff
@@ -813,6 +796,7 @@ def _ishangingEz(self):
                 hangingEz = np.kron(
                     np.ones(self.nCz, dtype=bool),
                     (
+                        # True * False = False
```
rowanc1: Thank you for reminding us of simple things. :) This is actually super helpful.

docs/conf.py
```diff
@@ -53,7 +53,7 @@
 
 # General information about the project.
 project = u'discretize'
-copyright = u'2013 - 2017, SimPEG Developers, http://simpeg.xyz'
+copyright = u'2013 - 2018, SimPEG Developers, http://simpeg.xyz'
```
rowanc1: Ehhhh!

discretize/CylMesh.py
lheagy: :) it is!

rowanc1:
I think this should for sure be a minor release.

docs/conf.py
lheagy: I know! Where does the time go...?

discretize/CylMesh.py
lheagy: yep, for sure

discretize/CylMesh.py
lheagy: :)

PR #38, opened 2017-03-07
Updating MeshIO, to address string replacement bug.

grosenkj:
Addressing simpeg/simpeg#548: "Import UBC mesh error"

Re-worked readCellLine function in readUBC to make the numpy array directly, skipping issues with double string matching errors.


rowanc1:
@grosenkj do you have a copy of the mesh that broke the assumptions? It would be good to include this in the tests so we don't break this again!

PR #41, opened 2017-03-26
Dot product Ax=y function in C. Needed for SimPEG.PF

fourndo:
A dot.product function in Cython to avoid memory issues using dot product and/or dot product of the transpose of a dense matrix. Needed for SimPEG.PF.
First stab at Cython, so this likely needs some love from and expert.


fourndo:
If this works, there are other functions in PF I would convert to C. All the trig functions in for-loop. 
Next step is adding OpenMP switch. Any of you guys tried?

jcapriot:
I've done openmp in cython before. However it is difficult to implement cross-platform, mostly because apples default clang compiler does not support openMP...

leouieda:
@fourndo what kind of memory issues are you having? 

The transpose in numpy doesn't copy memory, it's a different view of the same underlying memory block so it's cheap. 

The dot products in numpy are already done by underlying C functions, so how does this compare to `numpy.dot`? I'm asking because I've shot myself in the foot before spending weeks making a C function that in the end was slower than numpy.

leouieda:
I've also had the portability issues that @jcapriot mentioned with OpenMP in Fatiando. It worked perfectly on my linux machine but was a real pain to support in Mac and Windows. In the end, it wasn't really worth the extra pain that it caused.

Also, Anaconda ships with numpy and scipy linked against the MKL libraries, so if you have them a lot of matrix operations are already done multithreaded. They usually don't use all cores by default. That's a good thing because you're usually not running a single inversion and so there is always room for running multiple processes.

jcapriot:
@leouieda, I agree.

Have you looked at instead just using ``scipy.linalg.blas.dgemv`` ? It is a direct wrapper of the underlying blas functions that numpy's dot should be using, and you can definitely use that to avoid any copying of the data, (but it's usually a little less flexible).

leouieda:
@jcapriot I was thinking of doing something fancy and tying in to the scipy blas wrappers but I've been mostly focused on establishing an API that I like first. After that, I can dig down and do these kinds of optimizations without changing the API.

There are some big optimizations that can be done for inverse problems with selecting the proper blas or lapack routines, like setting the right options in [scipy.linalg.solve](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve). If the problem is properly regularized, it's safe to assume that the matrix is symmetric positive-definite so setting `sym_pos=True` might give decent speedups.

fourndo:
Thanks for taking the time gentlemen, really appreciated @leouieda @jcapriot .
So here is an interesting problem (see script attached).
If I run a memory_profile on this script, both the numpy.dot() and cython wrapper show the same memory consumption. But if you actually look at your memory in Task Manager (windows machine here) , you will see how the memory spikes constantly during the numpy.dot()

Turns out that the cython wrapper is factor 3x faster on my machine also.

Can you guys confirm this?


[DenseDotProduct.zip](https://github.com/simpeg/discretize/files/873859/DenseDotProduct.zip)

jcapriot:
What I can see about what is happening in that script you sent, is that ``A`` is ``np.float32``, and ``x`` is ``np.float64``. meaning in the dot product it will follow typical type promotion rules. Meaning ``A`` will get cast to ``np.float64`` (thus wasting a lot of time and making a duplicate). If you make ``A`` and ``x`` both the same ``dtype`` It substantially speeds up ``dot``.

leouieda:
That's a fair point about the dtypes. I tend to get sloppy with types because of Python but it can really get you when dealing with numerical operations. Another disadvantage of the Cython wrapper is that it won't work for different types easily (or things that aren't numpy arrays but follow the interface).

fourndo:
Yup you are right Joe. No memory jumps and faster if all float64. So if I want to work in float32 to save myself memory, I should convert x before calling numpy.dot?

fourndo:
Yep that works. Aight, I am killing this PR. Back to the numpy.dot()

leouieda:
@fourndo these things are tricky but it's the price we pay for the flexibility of dynamic typing. Glad it's solved with need of much coding :)

rowanc1:
Thanks @fourndo for doing this in the open. Learned a lot from the conversation!

Issue #26, opened 2017-02-04
Wheels for Windows


rowanc1:
Should consider building wheels for distribution without compilation.

@jcapriot do you have any idea how to do this? I think that we can create a wheel on appveyor, I will give it a first shot.

http://wheel.readthedocs.io/en/latest/
https://packaging.python.org/appveyor/

rowanc1:
See 400586c405e23f6923a9833e47404c11ddbde4b4.

rowanc1:
They are getting built, and saved. We may want 64bit? As well as python 2.7.

https://ci.appveyor.com/project/lheagy/discretize/build/1.0.50/artifacts

I will aim to upload these on the next release.

jcapriot:
32, and 64 both seem like good ideas, as well as 2.7 and 3.5.
The good thing about this is that we can package up the main simpeg branch as a universal python wheel (works on 2 and 3, and is pure python). So a single wheel file can be used to distribute all of the simpeg code.

jcapriot:
http://refi64.com/posts/using-appveyor-to-distribute-python-wheels.html

lheagy:
For distributing, does this mean we should do the entire pypi upload from appveyor? or can we upload to pypi from travis and then attach the wheels?

lheagy:
This looks like it might be what we need: https://pypi.python.org/pypi/appveyor-artifacts/1.0.2

lheagy:
Windows is still touchy and often Visual C++ issues arise. Perhaps we don't import TreeMesh stuff by default?? 

cc @fourndo, @rowanc1

PR #44, opened 2017-04-06
Distribution

lheagy:
- Build wheels on both linux and windows 
- include those in the deploy of discretize to pypi
- deployment to pypi is now automatic from appveyor and travis on a tag. 
- update travis testing to 3.6 being stable, 3.7 allowed to fail
image


lheagy:
This should close #26. If someone has a windows machine and can try `pip install discretize --pre` that would be much appreciated

rowanc1:
This is great @lheagy !

appveyor.yml
```diff
@@ -32,8 +49,22 @@ after_test:
...
   # 64-bit Python 3.3/3.4. And you need to use %PYTHON% to get the correct
   # interpreter
-  - python setup.py bdist_wheel
+  - python setup.py sdist bdist_wheel
 
 artifacts:
   # bdist_wheel puts your built wheel in the dist directory
   - path: dist\*
+
+deploy_script:
```
rowanc1: super awesome!

PR #47, opened 2017-04-13
Cyl Mesh Cleanup

lheagy:
## Cleanup on 3D cyl implementation
- use deflation matrices to build cyl mesh operators from tensor mesh operators

## Updates in TensorMesh
- pull out the stencil of the edgeCurl
- pull out stencil in faceDiv, cellGrad, nodalLaplacian
- separate averaging operators so they can be constructed component by component


discretize/CylMesh.py
```diff
@@ -338,6 +793,44 @@ def faceDivz(self):
...
+
+    #     if self.isSymmetric:
+    #         G1 = sp.kron(utils.speye(n[2]), ddxCellGrad(n[0], BC))
+    #     else:
+    #         G1 = self._deflationMatrix('Fx').T * utils.kron3(
+    #             utils.speye(n[2]), utils.speye(n[1]), ddxCellGrad(n[0], BC)
+    #         )
+    #     return G1
+
+    @property
```
lheagy: should raise a `NotImplementedError` for now - doesn't work yet

discretize/CylMesh.py
```diff
@@ -382,99 +875,12 @@ def edgeCurl(self):
...
-                O3 = utils.spzeros(self.nFz, self.nEz)
-
-                # Z contribution to the curl
-                Cz = sp.hstack((-Dr_z, Dt_z, O3))
-
                 self._edgeCurl = (
-                    utils.sdiag(1/A) *
-                    sp.vstack([Cr, Ct, Cz], format="csr") *
-                    utils.sdiag(E)
+                    utils.sdiag(1/self.area) *
```
lheagy: feels good to remove all of the above code :)

discretize/CylMesh.py
```diff
@@ -218,16 +398,276 @@ def vol(self):
...
+            deflate_y = np.zeros(self._ntNy, dtype=bool)
+            deflate_y[0] = True
+            deflateEx = np.nonzero(np.kron(
+                np.ones(self._ntNz, dtype=bool),
+                np.kron(
+                    deflate_y,
+                    np.ones(self.nCx, dtype=bool)
+                )
+            ))[0].tolist()
+            self.__hangingEx = dict(zip(
```
lheagy: there is a fair bit of changing of types here
(bool --> ints (indices) --> list --> list of tuples --> dict)

This was the way that was most convenient for me to work with - open to suggestions

Issue #48, opened 2017-04-22
cache the averaging matrices


lheagy:
right now, we use `sp.hstack` on every call to the averaging matrices to stack the contributions from the x, y, z components (if in 3D). This can be a slow operation for large meshes, so we should use lazy loading to cache the averaging matrices. 

See simpeg/simpeg#587

lheagy:
following from #49, 
"
Good point @ahartikainen. The operators would only need to be reset of the cell sizes changed. Initially, I was inclined to say that if you are changing the cell-sizes, you should just be creating a new mesh, and thus we would never need to clear the operators. However, for example, with the tree mesh, you may refine, run a test or something and then choose to refine or coarsen aspects of the mesh, in which case, we should clear all cached operators. 

cc @rowanc1. 

Maybe for naming it is 
```
mesh.clean()
``` 
? 
"

PR #49, opened 2017-04-22
cache the averaging operators

lheagy:
- cache `aveF2CC`, `aveF2CCV`, `aveE2CC`, `aveE2CCV` as per #48 
- question: do we want to be also caching the components (eg. `aveFx2CC`, ...)? Currently we are, but I think this might be unnecessary. The components are called once more to get the regularization operators constructed, but are then cached there, so it is really only one more call

### DIFF
--- a/discretize/DiffOperators.py
+++ b/discretize/DiffOperators.py
@@ -673,25 +673,34 @@ class DiffOperators(object):
     @property
     def aveF2CC(self):
         "Construct the averaging operator on cell faces to cell centers."
-        if(self.dim == 1):
-            return self.aveFx2CC
-        elif(self.dim == 2):
-            return (0.5)*sp.hstack((self.aveFx2CC, self.aveFy2CC),
-                                   format="csr")
-        elif(self.dim == 3):
-            return (1./3.)*sp.hstack((self.aveFx2CC, self.aveFy2CC,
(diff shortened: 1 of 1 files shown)


discretize/DiffOperators.py
```diff
@@ -702,11 +711,11 @@ def aveFx2CC(self):
 
         if getattr(self, '_aveFx2CC', None) is None:
             n = self.vnC
-            if(self.dim == 1):
+            if self.dim == 1:
```
rowanc1: Thanks for cleaning things up as well. :)

ahartikainen:
Should there be a function to reset all cached variables to None (if one edits the mesh)?

e.g.

    mesh.reset_cache()

lheagy:
Good point @ahartikainen. The operators would only need to be reset of the cell sizes changed. Initially, I was inclined to say that if you are changing the cell-sizes, you should just be creating a new mesh, and thus we would never need to clear the operators. However, for example, with the tree mesh, you may refine, run a test or something and then choose to refine or coarsen aspects of the mesh, in which case, we should clear all cached operators. 

cc @rowanc1. 

Maybe for naming it is 
```
mesh.clean()
``` 
?

lheagy:
I have included the discussion of `.clean()` in #48

Issue #51, opened 2017-04-27
Crash on install of SimPEG seems to be caused by line in discretize/setup.py


Rockpointgeo:
Hi all,

Trying to install SimPEG on my mac. Tried using pip in both python 2.7 and 3.5. I got the same error for both.

Collecting discretize (from SimPEG)
  Using cached discretize-0.1.4.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/kb/40gm7_gx27178vg4v7gswlxh0000gn/T/pip-build-yni1k5cp/discretize/setup.py", line 9, in <module>
        from numpy.distutils.core import setup
    ImportError: No module named 'numpy'

Looking at the setup.py file I see that on line 9 it says

from numpy.distutils.core import setup

By no means a python expert but looking at other setup.py files and they mainly seem to have the line

from distutils.core import setup

Could this be the problem? If not back to the drawing board. Any help would be appreciated.

Thanks
Sean

lheagy:
hi @Rockpointgeo, do you have python installed through anaconda?

It look like python / pypi might not be pathed properly on your machine. If you open a terminal and tun `which python` what is the output?

lheagy:
@rowanc1, @jcapriot: do you guys have any thoughts / comments on best practices for creating setup.py files for a package? That is something I could use some pointers on (and in particular when helping troubleshoot instillation)

jcapriot:
My first guess looking at the error is to make sure numpy is installed properly. The setup.py files do use numpy's version of distutils, but obviously that requires numpy to be installed first. So I'd start by checking that.

lheagy:
I think we got it sorted for sean's case (I will let him confirm!). Is the main motivation for using the numpy distutils (rather than the core distutils) for cython?

Rockpointgeo:
Lindsey gave me some help on this. I tried a few things.

1) If I install python using Anaconda during installation many packages are installed including numpy. Then if I install SimPEG via pip there are no errors.

2) If I install python using something else (pyenv via Homebrew for example) no packages are installed during the python installation. If I install SimPEG via pip I get the error.

3) If I install python using something else (pyenv via Homebrew for example) then install the packages listed as dependencies (numpy, scipy, matplotlib, and cython). If I install SimPEG via pip everything is ok.

[...]

PR #56, opened 2017-05-16
setup.py

lheagy:
- better error handling on the import of numpy in setup.py as per #51
- if cython interputils not built - tell user to run `python setup.py install` (not `python setup.py build_ext --inplace`) @rowanc1 or @jcapriot: could one of you double-check me on this? Thanks!

### DIFF
--- a/discretize/utils/interputils.py
+++ b/discretize/utils/interputils.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     print("""Interpolation will not work, use setup.py to compile the cython:
 
-        python setup.py build_ext --inplace""")
+        python setup.py install""")
     _interpCython = False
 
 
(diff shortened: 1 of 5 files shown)


discretize/utils/interputils.py
```diff
@@ -13,7 +13,7 @@
 except ImportError:
     print("""Interpolation will not work, use setup.py to compile the cython:
 
-        python setup.py build_ext --inplace""")
+        python setup.py install""")
```
lheagy: @rowanc1: can you double check me on this?

jcapriot: The only way interputils would fail to import is if it wasn't installed correctly, so technically this would fix that problem.

However if interputils would fail to build, the install command would not happen, thus you wouldn't be able to even import discretize (because it isn't installed), from the python shell outside the source directory.

Essentially this problem shouldn't come up in normal practice as it would throw the error when it is being installed, and not when it is imported.

lheagy: Quite true, thanks @jcapriot. I will instead add a warning like this in the root init (if you installed from source and didn't run setup.py, that would break)

rowanc1: Thanks @jcapriot and @lheagy

Issue #60, opened 2017-06-12
utils.spzeros efficiency


lheagy:
I think we want to use `sp.dia_matrix` instead of `sp.csr_matrix` for `utils.spzeros` (2x speedup) 

![image](https://user-images.githubusercontent.com/6361812/27048362-4e962472-4f60-11e7-8aa0-ad0036027167.png)

lheagy:
and if the matrix isn't square, it is still faster
![image](https://user-images.githubusercontent.com/6361812/27049479-599b8f0c-4f64-11e7-93ee-42db93d9f2a4.png)

jcapriot:
That makes a lot of sense, considering how the csc/csr matrices are stored, an empty diagonal matrix doesn't require any storage space, whereas the csc/csr matrices still require offset storage which is just an array of zeros the length of columns/rows.

lheagy:
closed with #61

PR #61, opened 2017-06-12
Ref/utils spzeros

lheagy:
use `dia_matrix` instead of `csr_matrix` for spzeros (see #60)

### DIFF
--- a/discretize/utils/matutils.py
+++ b/discretize/utils/matutils.py
@@ -413,6 +413,7 @@ def invPropertyTensor(M, tensor, returnMatrix=False):
 class Zero(object):
 
     __numpy_ufunc__ = True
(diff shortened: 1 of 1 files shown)


lheagy:
The current travis failure is due to a numpy update and the `Utils.Zero()` ...

discretize/utils/matutils.py
```diff
@@ -488,6 +489,7 @@ def T(self):
 class Identity(object):
 
     __numpy_ufunc__ = True
+    __array_ufunc__ = None
```
rowanc1: good catch!

lheagy: thanks rowan!

Issue #66, opened 2017-06-28
Extract core mesh from TreeMesh


rowanc1:
The plotting for an octree is insanely slow. There is also some times that you want to work with a core mesh as a tensor mesh. This requires you to extract pieces and use those.

![image](https://user-images.githubusercontent.com/913249/27615223-edbc8140-5b63-11e7-8b5d-00b77cf11820.png)

A quick sketch of the algorithm in 2D is something like:

```python
# find the first cell of the tree mesh
x0_tensor = MT.gridCC[0, :]
for c in M:
    if np.allclose(c.center, x0_tensor):
        first_cell = c
        break
print(first_cell.center)
print(first_cell._pointer)
ptr = first_cell._pointer

index = np.zeros(MT.vnC, dtype=int)

for ii, nx in enumerate(range(MT.nCx)):
    for jj, ny in enumerate(range(MT.nCy)):
        # if you are not in the last level of the tree mesh 
        # you will have to skip up by that number
        # The last entry of the pointer refers to your level
        ptr_ij = [ptr[0] + nx, ptr[1] + ny, ptr[2]]
        index[ii, jj] = M._cc2i[M._index(ptr_ij)]

index = discretize.utils.mkvc(index)
```

And then you can plot it:
```python
fig, axs = plt.subplots(1, 2, figsize=(15,8))
v = np.sin(M.gridCC[:, 0] * 2.0 * np.pi) * np.sin(M.gridCC[:, 1] * 2.0 * np.pi) + M.gridCC[:, 1]
M.plotImage(v, ax=axs[0])
axs[0].plot(MT.gridCC[:,0], MT.gridCC[:,1], 'kx')

MT.plotImage(v[index], ax=axs[1])
```

Would need a bunch of error checking, and extensions to 3D to make this robust.

cc @micmitch

rowanc1:
A gotcha would be the pointer code if you are not on the lowest level.
Should error in the `M._cc2i` lookup if the cell does not exist.
you will have to step by the gap in `ptr_ij = [ptr[0] + nx, ptr[1] + ny, ptr[2]]` line
You could change the `enumerate(range(MT.nCy*level_from_bottom)[::level_from_bottom])` as a quick hack. 
i.e. level from bottom in this case is 1

domfournier:
I am sure @jcapriot  will have a better trick, but this would avoid looping and using built-in discretize functions:
`
core_cells = mesh.cell_volumes == mesh.cell_volumes.min()
[...]

PR #67, opened 2017-07-03
Extract Core Mesh

lheagy:
- port over fixes that were made in SimPEG extract core mesh

### DIFF
--- a/discretize/utils/meshutils.py
+++ b/discretize/utils/meshutils.py
@@ -173,7 +173,8 @@ def closestPoints(mesh, pts, gridLoc='CC'):
 
 
 def ExtractCoreMesh(xyzlim, mesh, meshType='tensor'):
-    """Extracts Core Mesh from Global mesh
+    """
+    Extracts Core Mesh from Global mesh
 
     :param numpy.ndarray xyzlim: 2D array [ndim x 2]
     :param BaseMesh mesh: The mesh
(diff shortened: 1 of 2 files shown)


discretize/utils/meshutils.py
```diff
@@ -173,17 +173,16 @@ def closestPoints(mesh, pts, gridLoc='CC'):
...
 
     This function ouputs::
 
         - actind: corresponding boolean index from global to core
-        - meshcore: core mesh
-
-    Warning: 1D and 2D has not been tested
+        - meshcore: core sdiscretize mesh
     """
     import discretize
```
thast: The function works again. Thanks.
Would moving that `import` outside of the function improve it too?

thast: Also, we can add a test later this week to improve codecov coverage. I have some 2D and 3D examples we can use.

rowanc1: Examples and tests sound awesome.

I think we might get into some import issues if we take `discretize` outside of the function. It is definitely better practice, though. I think ideally this would be a method on the mesh?

```python
core_mesh = mesh.core_mesh(lims)
```

I think for now just adding the tests would be good enough. Maybe an `assert` statement that doesn't let you do anything else other than `meshType='tensor'`? There is some of this started on for the `TreeMesh` on an issue. 

See #66.

lheagy: @thast: adding a test would be awesome! I will hold off on merging this until we add one. Would you be game to do that??

@rowanc1: agreed, later on, we should think about making this a method on the mesh, but for now, I think the scope of this pr is to get the util cleaned up so that we are at least back to baseline functionality.

tests/base/test_utils.py
```diff
@@ -381,5 +382,47 @@ def test_both(self):
         assert o*z + o == 1
         assert o-z == 1
 
+
+class TestMeshUtils(unittest.TestCase):
```
lheagy: Thanks @thast!!

PR #70, opened 2017-09-01
Ref/properties

lheagy:
# Implement properties in discretize

Use properties to define the properties of a class that are necessary to completely define the object. This allows us to serialize, save and load meshes (also copy!)

image

## todos
- [x] test on cyl mesh
- [x] test on tree mesh
- [x] test on curvi mesh

@rowanc1, @fwkoch: if you guys would be willing to do a review, I would appreciate another set of eyes

## Related issues 
- private properties: aranzgeo/properties#191
- arrays with arbitrary shape: aranzgeo/properties#192


README.rst
```diff
@@ -21,9 +21,6 @@ discretize
     :target: https://www.codacy.com/app/lindseyheagy/discretize?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=simpeg/discretize&amp;utm_campaign=Badge_Grade
     :alt: codacy status
 
-.. image:: https://www.quantifiedcode.com/api/v1/project/97126041a7e7400fbf1c5555bed514dc/badge.svg
```
lheagy: we weren't really using this and they discontinued it

discretize/BaseMesh.py
```diff
@@ -290,11 +306,33 @@ def projectEdgeVector(self, eV):
...
+        Make a copy of the current mesh
+        """
+        return properties.copy(self)
+
 
 class BaseRectangularMesh(BaseMesh):
     """BaseRectangularMesh"""
     def __init__(self, n, x0=None):
-        BaseMesh.__init__(self, n, x0)
+        BaseMesh.__init__(self, n, x0=x0)
```
lheagy: this isn't strictly necessary. just for clarity

discretize/BaseMesh.py
```diff
@@ -590,16 +628,34 @@ def r(self, x, xType='CC', outType='CC', format='V'):
             eX, eY, eZ = r(edgeVector, 'E', 'E', 'V')
         """
 
-        assert (type(x) == list or isinstance(x, np.ndarray)), "x must be either a list or a ndarray"
-        assert xType in ['CC', 'N', 'F', 'Fx', 'Fy', 'Fz', 'E', 'Ex', 'Ey', 'Ez'], "xType must be either 'CC', 'N', 'F', 'Fx', 'Fy', 'Fz', 'E', 'Ex', 'Ey', or 'Ez'"
-        assert outType in ['CC', 'N', 'F', 'Fx', 'Fy', 'Fz', 'E', 'Ex', 'Ey', 'Ez'], "outType must be either 'CC', 'N', 'F', Fx', 'Fy', 'Fz', 'E', 'Ex', 'Ey', or 'Ez'"
+        allowed_xType = [
```
lheagy: whitespace cleanup

discretize/CurvilinearMesh.py
```diff
@@ -25,7 +27,62 @@ def normalize3D(x):
     return x/np.kron(np.ones((1, 3)), utils.mkvc(length3D(x), 2))
 
 
-class CurvilinearMesh(BaseRectangularMesh, DiffOperators, InnerProducts, CurviView):
+class Array(properties.Array):
```
lheagy: This is the implementation of an array that doesn't care about shape (I allow shape to be `'*'` which could be a 1D, 2D or 3D array). See discussion on aranzgeo/properties#192

discretize/TreeMesh.py
```diff
@@ -98,23 +99,48 @@ class TreeMesh(BaseTensorMesh, InnerProducts, TreeMeshIO):
 
     _meshType = 'TREE'
 
-    def __init__(self, h, x0=None, levels=None):
+    levels = properties.Integer(
+        "discretization level",
+        min=0
+    )
+
+    cells = properties.Set(
```
lheagy: I would prefer to keep this a private variable `_cells`. See discussion on aranzgeo/properties#191

discretize/TreeUtils.c
```diff
@@ -1,8 +1,8 @@
-/* Generated by Cython 0.24.1 */
+/* Generated by Cython 0.25.2 */
```
lheagy: If you prefer, I can remove this. Updated cython cc @rowanc1

tests/base/test_tensor.py
```diff
@@ -19,7 +19,6 @@ def setUp(self):
     def test_vectorN_2D(self):
         testNx = np.array([3, 4, 5, 6])
         testNy = np.array([5, 6, 8])
-
```
lheagy: just whitespace in this file

discretize/BaseMesh.py
```diff
@@ -1,41 +1,57 @@
...
-    :param numpy.array n: (or list) number of cells in each direction (dim, )
-    :param numpy.array x0: (or list) Origin of the mesh (dim, )
-
     """
 
+    # Properties
+    n = properties.Array(
+        "number of cells in each direction (dim, )",
+        dtype=int,
+        required=True,
```
rowanc1: Do we want `min_length` or `max_length` here?

http://propertiespy.readthedocs.io/en/latest/content/container.html#properties.List

discretize/BaseMesh.py
```diff
@@ -1,41 +1,57 @@
...
-        return self._x0
+            self.x0 = np.zeros(len(self.n))
+        else:
+            self.x0 = x0
+
+    # Validators
+    @properties.validator('n')
+    def check_n_shape(self, change):
+        change['value'] = np.array(change['value'], dtype=int).ravel()
+        if len(change['value']) > 3:
```
rowanc1: I think this can be solved by min/max length in the list property.

discretize/BaseMesh.py
rowanc1: Actually, that is only for lists. Perhaps we could change properties to have a `set` of tuples for defining the shape?

`shape={(1,), (2,), (3,)}`??

discretize/BaseMesh.py
```diff
@@ -1,41 +1,57 @@
...
+    BaseMesh does all the counting you don't want to do.
     BaseMesh should be inherited by meshes with a regular structure.
-
-    :param numpy.array n: (or list) number of cells in each direction (dim, )
-    :param numpy.array x0: (or list) Origin of the mesh (dim, )
-
     """
 
+    # Properties
+    n = properties.Array(
```
rowanc1: Can we change this later? Having it as a public property seems to suggest that we can change it....

discretize/CurvilinearMesh.py
```diff
@@ -41,28 +115,48 @@ class CurvilinearMesh(BaseRectangularMesh, DiffOperators, InnerProducts, CurviVi
...
-    def __init__(self, nodes):
-        assert type(nodes) == list, ("'nodes' variable must be a list of "
-                                     "np.ndarray")
-        assert len(nodes) > 1, "len(node) must be greater than 1"
+    nodes = properties.List(
+        "List of arrays describing the node locations",
+        Array(
+            "node locations in a dimension",
+            shape='*'
+        ),
```
rowanc1: min=2 and max=3 length on the list?

discretize/BaseMesh.py
lheagy: see aranzgeo/properties#191

discretize/BaseMesh.py
lheagy: not sure if we need something that complex? this is just 3 numbers (at most), nCx, nCy, nCz

lheagy: That would be handy

discretize/BaseMesh.py
lheagy: This isn't a list (it is an array). Which I don't think has a min/max length

discretize/BaseMesh.py
lheagy: agreed though - ideally, this should remain a private variable

discretize/BaseMesh.py
```diff
@@ -1,41 +1,57 @@
...
+    def check_n_shape(self, change):
+        change['value'] = np.array(change['value'], dtype=int).ravel()
+        if len(change['value']) > 3:
+            raise Exception(
+                "Dimensions of {}, which is higher than 3 are not "
+                "supported".format(change['value'])
+            )
+
+    @properties.validator('x0')
+    def check_x0_vs_n(self, change):
```
fwkoch: A couple picky thoughts on this:
- `self.n` _could_ be `None` if someone deletes it or sets it to undefined. `len(None)` would give an ugly error. Possibly worth adding an extra sanity check for that?
- Also, possibly worth running this check in the other direction (ie you validate `x0` vs `n`, but not `n` vs `x0`) - may come up depending on how someone is using this.

discretize/BaseMesh.py
```diff
@@ -44,7 +60,7 @@ def dim(self):
         :rtype: int
         :return: dim
         """
-        return self._dim
+        return len(self.n)
```
fwkoch: For all of these, actually, possibly worth checking if things (`self.n`, `self.dims`, etc.) are `None`.

discretize/CurvilinearMesh.py
fwkoch: Union property could work well - see my comment on aranzgeo/properties#192

discretize/CurvilinearMesh.py
```diff
@@ -41,28 +115,48 @@ class CurvilinearMesh(BaseRectangularMesh, DiffOperators, InnerProducts, CurviVi
...
+    )
 
-        for i, nodes_i in enumerate(nodes):
-            assert isinstance(nodes_i, np.ndarray), ("nodes[{0:d}] is not a"
-                                                     "numpy array.".format(i))
-            assert nodes_i.shape == nodes[0].shape, ("nodes[{0:d}] is not the "
-                                                     "same shape as nodes[0]"
-                                                     .format(i))
+    def __init__(self, nodes=nodes, **kwargs):
```
fwkoch: Looks like you are using `nodes`, the `properties.List`, as the default value...? Seems fishy... though I may be missing something.

discretize/CylMesh.py
```diff
@@ -29,17 +30,39 @@ class CylMesh(BaseTensorMesh, BaseRectangularMesh, InnerProducts, CylView):
...
+
+        # assert len(self.cartesianOrigin) == self.dim, (
+        #     "cartesianOrigin must be the same length as the dimension"
+        #     " of the mesh."
+        # )
+        # self.cartesianOrigin = np.array(self.cartesianOrigin, dtype=float)
+
+    @properties.validator('cartesianOrigin')
+    def check_cartesian_origin_shape(self, change):
+        change['value'] = np.array(change['value'], dtype=float).ravel()
```
fwkoch: This is cool! I like seeing that you are using the validator to (possibly) change the value.

discretize/TensorMesh.py
```diff
@@ -18,25 +19,58 @@ class BaseTensorMesh(BaseMesh):
...
+                    "h[{0:d}] is not a numpy array.".format(i)
+                )
+                assert len(h_i.shape) == 1, (
+                    "h[{0:d}] must be a 1D numpy array.".format(i)
+                )
+                h[i] = h_i[:]  # make a copy.
+
+        else:
+            assert 'h' in kwargs.keys(), 'cell widths must be provided'
+            h = kwargs.pop('h')
```
fwkoch: This is a little funny - on first impulse, my question was "Why does `h` get special treatment as an arg but not as a kwarg?" Looking into it further, though, as long as h is provided as an arg or a kwarg, it will set `h_in`. So really, this line will never be reached; if the `else` clause is reached, the `assert` will fail (I think at least).

Silly example:
image

discretize/TensorMesh.py
```diff
@@ -55,38 +89,35 @@ def __init__(self, h_in, x0_in=None):
...
+            n = kwargs.pop('n')
+            assert (n == np.array([x.size for x in h])).all(), (
+                "Dimension mismatch. The provided n doesn't "
             )
-        else:
-            BaseMesh.__init__(self, np.array([x.size for x in h]), x0)
 
-        # Ensure h contains 1D vectors
-        self._h = [utils.mkvc(x.astype(float)) for x in h]
+        n = np.array([x.size for x in h])
```
fwkoch: Should this be `else`?

discretize/utils/meshutils.py
```diff
@@ -12,6 +14,22 @@
     num_types = [int, float]
 
 
+def load_mesh(filename):
+    """
+    Open a json file and load the mesh into the target class
+
+    As long as there are no namespace conflicts, the target __class__
```
fwkoch: If you want, you can narrow down your namespace a little by overriding the `_REGISTRY` on your `BaseMesh` class (ie just add `_REGISTRY = {}` to the class). Then you could say `data = BaseMesh.deserialize(...)` and it will only look at classes that inherit from `BaseMesh` (and therefore are on the BaseMesh registry, rather than the HasProperties registry).

discretize/TensorMesh.py
lheagy: It is a bit odd, and maybe you have a suggestion for me on how this might be made more transparent?

The reason for this is to support two uses: 
1. creating the mesh with `h, x0` as args (or `h_in`). This is currently how most people work with the API
2. the reason for having `h` also allowed as a kwarg is so that I can create a re-create a mesh from a serialized mesh

image

where `load_mesh` deserializes the mesh (https://github.com/simpeg/discretize/blob/ref/properties/discretize/utils/meshutils.py#L17)

```python
def load_mesh(filename):
    """
    Open a json file and load the mesh into the target class
    As long as there are no namespace conflicts, the target __class__
    will be stored on the properties.HasProperties registry and may be
    fetched from there.
    :param str filename: name of file to read in
    """
    with open(filename, 'r') as outfile:
        jsondict = json.load(outfile)
        data = properties.HasProperties.deserialize(jsondict, trusted=True)
    return data
``` 

and for this I am under the impression that all properties in the serialized json needed to be able to be read in as kwargs.

discretize/TreeMesh.py
```diff
@@ -98,23 +99,48 @@ class TreeMesh(BaseTensorMesh, InnerProducts, TreeMeshIO):
...
+        "cells in the tree mesh",
+        properties.Integer(
+            "cell number",
+            min=0
+        ),
+        default=set()
```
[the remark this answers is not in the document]
lheagy: ah, thanks

discretize/utils/meshutils.py
lheagy: Ok, cool, thanks!

discretize/TensorMesh.py
fwkoch: I see the need for h as both arg and kwarg - I think that is still covered.

This exposes a bit of a subtlety with how Python treats args/kwargs. If a named arg variable shows up anywhere in the kwargs (and is not provided in the args) it is assigned to the associated arg (and that variable does not show up in `kwargs` internal to the function).

If you play around with this:

```python
class Blah(object):
    
    def __init__(self, h=None, x0=None, **kwargs):
        print(h)
        print(x0)
        print(kwargs)
```

You will find for all of these, `kwargs = {'something_else': 2}`:

```python
Blah(0, 1, something_else=2)
Blah(h=0, x0=1, something_else=2)
Blah(something_else=2, x0=1, h=0)
```
This errors (h assigned twice):
```python
Blah(0, x0=1, something_else=2, h=3)
```

So bottom line, even if it is coming in as a deserialized dict, the argument will still be moved to the right place. Does that make sense...?

discretize/CurvilinearMesh.py
lheagy: This is an array, so we can't, but I will implement the Union Property as suggested by @fwkoch

discretize/CurvilinearMesh.py
lheagy: thanks, this is a typo.

lheagy:
Properties can now be private (see aranzgeo/properties#194, thanks @fwkoch!!), so...  
- on the BaseMesh `n` is a private property (`_n`)
- on the TreeMesh, `levels` and `cells` are private  (`_levels` and `_cells`)

and Arrays can now be given a set for the shape (see aranzgeo/properties#195), so 
- on the CurviMesh, `nodes` is a list of arrays.

I am happy with the state of this pr at this point, @rowanc1, if you are as well, I will do a beta release and merge to dev.

discretize/BaseMesh.py
```diff
@@ -290,11 +351,33 @@ def projectEdgeVector(self, eV):
         ), 'eV must be an ndarray of shape (nE x dim)'
         return np.sum(eV*self.tangents, 1)
 
+    def save(self, filename='mesh.json', verbose=False):
```
rowanc1: This might want to be on a base class for simpeg in future (or in the utils of discretize)? It looks like a pretty generic function.

discretize/BaseMesh.py
```diff
@@ -590,44 +673,69 @@ def r(self, x, xType='CC', outType='CC', format='V'):
...
         assert format in ['M', 'V'], "format must be either 'M' or 'V'"
-        assert outType[:len(xType)] == xType, "You cannot change types when reshaping."
-        assert xType in outType, 'You cannot change type of components.'
+        assert outType[:len(xType)] == xType, (
+            "You cannot change types when reshaping."
+        )
+        assert xType in outType, "You cannot change type of components."
+
         if type(x) == list:
             for i, xi in enumerate(x):
```
rowanc1: It looks like this never gets hit by tests. Did it before?

discretize/CurvilinearMesh.py
```diff
@@ -41,37 +45,60 @@ class CurvilinearMesh(BaseRectangularMesh, DiffOperators, InnerProducts, CurviVi
...
 
-    def __init__(self, nodes):
-        assert type(nodes) == list, ("'nodes' variable must be a list of "
-                                     "np.ndarray")
-        assert len(nodes) > 1, "len(node) must be greater than 1"
+    nodes = properties.List(
+        "List of arrays describing the node locations",
+        prop=properties.Array(
+            "node locations in an n-dimensional array",
+            shape={('*', '*'), ('*', '*', '*')}
```
rowanc1: Cool!

discretize/BaseMesh.py
lheagy: Agreed. Once we have injected properties elsewhere in SimPEG we can abstract this up.

discretize/BaseMesh.py
lheagy: The only thing I changed above is whitespace. This particular line has not changed at all in this pr

PR #72, opened 2017-09-16
Tree mesh cell grad

fourndo:
Cell grad operators for TreeMesh inversion
-----------------------------------------------
- First pass implementation for stencil operators
Initial implementation without hanging faces
image

- Second implementation with hanging faces
image

**Smoother gradients on Octree change.**

- Create gradient operators with dimensions for PDE problems.
- Add unit tests for operators. Thanks @micmitch and @sgkang 

Solve DC problem in mine tunnels.
Tree mesh examples will follow shortly on SimPEG ... stay tuned.
image

PR #73, opened 2017-09-21
discretize 0.1.10

lheagy:
- [x] **properties**: injection of properties into discretize from #70. This allows meshes to be serialized, saved, and loaded. 
- [ ] **Tree Mesh Cell Grad** (@fourndo) : implement a cell gradient for the Tree mesh from #72 
- [ ] **3D Cyl Mesh**: first implementation of a 3D cyl mesh from #36. This also makes the cell grad stencils properties rather than functions


lheagy:
I would like to change the scope of this and bring it in as-is. The cyl mesh changes and Tree Mesh cell grad can be separate releases (and directly onto master). @fourndo, does this work with you?

PR #77, opened 2017-10-02
GridH function

dccowan:
This function works for TensorMesh and TreeMesh types. Much like gridCC, gridH returns [hx,hy,hz] as a Nx3 numpy array. This function is important for multiple potential fields problems (Mag, VRM, grav) in 3d.


discretize/TensorMesh.py
```diff
@@ -160,6 +160,24 @@ def gridN(self):
         return self._getTensorGrid('N')
 
     @property
+    def gridH(self):
+        """Cell widths in grid."""
+        H = self.h
+        if self.dim == 1:
+            return H
+        if self.dim == 2:
+            H = np.meshgrid(H[0],H[1])
```
lheagy: could you please install a pep8 linter? (see for example: http://docs.simpeg.xyz/content/api_core/api_practices.html#style). This really helps with readability and maintenance in the long run.

eg. there should be a space after the comma
```
H = np.meshgrid(H[0], H[1])
```

discretize/TensorMesh.py
```diff
@@ -160,6 +160,24 @@ def gridN(self):
         return self._getTensorGrid('N')
 
     @property
+    def gridH(self):
```
lheagy: @dccowan: do you have an example of where this is used? Some context would be helpful. 

A couple comments: this isn't really a `grid` in the same sense as we are using the word elsewhere in the codebase, so we should think about the naming here. `grid` is really used to denote locations in space (spatial locations of cell centers, nodes, faces, etc), whereas this is a tensor representation of the cell widths. 

I am also a bit hesitant to add something to the class that is really a view of information that we already have accessible through `mesh.h`. This is where having an example and a bit of context would be quite helpful.

discretize/TensorMesh.py
```diff
@@ -160,6 +160,24 @@ def gridN(self):
...
+        H = self.h
+        if self.dim == 1:
+            return H
+        if self.dim == 2:
+            H = np.meshgrid(H[0],H[1])
+            hx = np.reshape(H[0],self.nC)
+            hy = np.reshape(H[1],self.nC)
+            return np.c_[hx,hy]
+        elif self.dim == 3:
+            H = np.meshgrid(H[0],H[1],H[2])
```
rowanc1: I also think that there may be some problems with the ordering on this `meshgrid`. We have `utils.ndgrid` that does things how we expect them in the vectorized form (`'F'` order). A test of uneven grids would be helpful (plotting even to make sure...).

https://github.com/simpeg/discretize/blob/dev/discretize/utils/matutils.py#L101

lheagy:
What about a name of `cell_widths` for this? or `h_gridded`? maybe the latter makes the most sense, it returns a gridded representation of the cell widths.  

cc @rowanc1, @fourndo

lheagy:
or `hx_by_cell`?

discretize/TreeMesh.py
```diff
@@ -728,6 +728,15 @@ def gridN(self):
         R = self._deflationMatrix('N', withHanging=False)
         return R.T * self._gridN + np.repeat([self.x0],self.nN,axis=0)
 
+    @property
+    def gridH(self):
+        if getattr(self, '_gridH', None) is None:
```
lheagy: This one still isn't tested, would you mind including a test for the tree mesh?

discretize/TensorMesh.py
```diff
@@ -159,6 +159,22 @@ def gridN(self):
...
+            return self.h
+        if self.dim == 2:
+            hx = np.kron(np.ones(self.nCy), self.hx)
+            hy = np.kron(self.hy, np.ones(self.nCx))
+            return np.c_[hx, hy]
+        elif self.dim == 3:
+            hx = np.kron(np.ones(self.nCy*self.nCz), self.hx)
+            hy = np.kron(np.ones(self.nCz), np.kron(self.hy, np.ones(self.nCx)))
+            hz = np.kron(self.hz, np.ones(self.nCx*self.nCy))
+            return np.c_[hx, hy, hz]
```
lheagy: are there cases where you want to access these individually? (eg. hx, hy, hz) or are you always using all of them when asking for it?

discretize/TreeMesh.py
```diff
@@ -728,6 +728,15 @@ def gridN(self):
         R = self._deflationMatrix('N', withHanging=False)
         return R.T * self._gridN + np.repeat([self.x0],self.nN,axis=0)
 
+    @property
+    def h_gridded(self):
```
lheagy: could you please add a docstring to this?

discretize/TensorMesh.py
```diff
@@ -159,6 +159,22 @@ def gridN(self):
         """Nodal grid."""
         return self._getTensorGrid('N')
 
+    @property
+    def h_gridded(self):
+        """Dimensions of all cells in order."""
```
lheagy: would you mind including dimensions in the docstring? eg
```
returns an nC x dim array of the cell dimensions (eg. [hx, hy, hz] for a 3D mesh) 
```

discretize/TreeMesh.py
dccowan: Is this done through simpeg or discretize repository or both?

lheagy: this is done in the code
```
"""
Docstring for h_gridded
"""
```

discretize/TensorMesh.py
dccowan: Oh i see. Once of those

discretize/TensorMesh.py
```diff
@@ -159,6 +159,23 @@ def gridN(self):
         """Nodal grid."""
         return self._getTensorGrid('N')
 
+    @property
+    def h_gridded(self):
+        """Returns an M by N numpy array with the widths of all cells in order.
```
lheagy: sorry to be picky, but for docstrings that span multiple lines, would you mind using the convention of opening the docstring on one line and closing it on a separate line. 
```
"""
Returns an (nC, dim) numpy array with the widths of all cells in order
"""
```

discretize/TreeMesh.py
```diff
@@ -715,6 +715,12 @@ def balance(self, recursive=True, cells=None, verbose=False, _inRecursion=False)
 
     @property
     def gridCC(self):
+        """
```
rowanc1: Thanks for the doc strings!

discretize/TensorMesh.py
```diff
@@ -159,6 +159,24 @@ def gridN(self):
...
         return self._getTensorGrid('N')
 
+    @property
+    def h_gridded(self):
+        """
+        Returns an (nC, dim) numpy array with the widths of all cells in order
+        """
+
+        if self.dim == 1:
+            return self.h
```
rowanc1: I believe this is len `(nC,)` rather than what is stated in the doc string. Probably what is wanted, but a bit confusing. Could use the `mkvc(self.h, 1)` I think? to add a blank axis?

discretize/TreeMesh.py
```diff
@@ -724,10 +730,28 @@ def gridCC(self):
...
         R = self._deflationMatrix('N', withHanging=False)
         return R.T * self._gridN + np.repeat([self.x0],self.nN,axis=0)
 
+    @property
+    def h_gridded(self):
+        """
+        Returns an (nC, dim) numpy array with the widths of all cells in order
+        """
+
+        if getattr(self, '_h_gridded', None) is None:
```
rowanc1: Is this deleted when the mesh is changed?
I would need to look at the code to point you to exactly where that is done, but if the mesh is refined/coarsened then this needs to be deleted.

lheagy: nice catch, no this would still be stashed if it had been created

dccowan: It is/isn't stashed in the exact same way the gridCC is/isn't stashed. I'll leave that decision to the higher ups.

lheagy: @dccowan: this is the issue that @rowanc1 is referring to. You are using "lazy loading" here. The first time `h_gridded` is asked for, it is created and stored as a private variable (`_h_gridded`). It the mesh is refined, right now `_h_gridded` is not deleted

image

discretize/TreeMesh.py
```diff
@@ -170,7 +170,7 @@ def __dirty__(self, val):
             '_faceDiv', '_edgeCurl', '_nodalGrad',
             '_aveFx2CC', '_aveFy2CC', '_aveFz2CC', '_aveF2CC', '_aveF2CCV',
             '_aveEx2CC', '_aveEy2CC', '_aveEz2CC', '_aveE2CC', '_aveE2CCV',
-            '_aveN2CC',
+            '_aveN2CC', '_h_gridded'
```
lheagy: @dccowan, @rowanc1: this now makes sure that `_h_gridded` is cleared on mesh updates. There is also a test.

discretize/TreeMesh.py
lheagy: In the last commit, this is now cleaned up.

dccowan:
Some stuff still isn't passing. Not really sure why. At first glance, it looks like stuff I don't really touch.

lheagy:
@dccowan: I think it is an upstream issue with the current anaconda distribution... It is going to take a bit of hunting.

lheagy:
@rowanc1: did you want to have a look at this before we bring it in?

PR #84, opened 2018-01-03
add IO functionalities to write DCIP2D mesh and models

thast:
- the `MeshIO` functions `writeUBC` and `writeModelUBC` are now compatible with 2D meshes and models.

- We now make sure that mesh and model get saved at the same place by giving the option to specify the folder outside of the filename (before the folder was given by the `dict.keys`, not very practical)

- The upgrade is backward compatible


discretize/MeshIO.py
```diff
@@ -300,7 +300,56 @@ def _toVTRObj(mesh, models=None):
             vtkObj.GetCellData().SetActiveScalars(models.keys()[0])
         return vtkObj
 
-    def readModelUBC(mesh, fileName):
+    def _readModelUBC_2D(mesh, fileName):
+        """
```
lheagy: Thanks for this @thast!

discretize/MeshIO.py
```diff
@@ -300,7 +300,56 @@ def _toVTRObj(mesh, models=None):
...
+
+            for ii in range(len(obsfile)-1):
+                mm = np.array(obsfile[ii+1].split(), dtype=float)
+                model[:, ii] = mm
+
+            model = model[:, ::-1]
+
+        else:
+
+            if len(obsfile[1:]) == 1:
```
lheagy: would you mind adding a test that hits these lines?

tests/base/2d_2cyl_model
```diff
@@ -0,0 +1,48 @@
+135 47
```
lheagy: @thast: is this for testing, if you slack me the file, I will upload and send you a url where you can download it so it is not in the core repo

tests/base/ubc_DC2D_tensor_mesh.msh
```diff
@@ -1,35 +1,26 @@
...
-2700 1
-2750 1
-2800 1
-2900 1
-3050 1
-3250 1
-3550 1
-3950 1
-4550 1
+!comment line
```
lheagy: same here, I am happy to upload this

tests/base/2d_2cyl_model
thast: @lheagy : Thanks for pointing it out. The test actually writes and then read back the testing files here (so we test both input and output). 
I will reproduce what we do in SimPEG and create a temporary folder that we then delete, so no storing will be necessary.

discretize/MeshIO.py
thast: we can actually make this function much simpler. Will do.

discretize/MeshIO.py
```diff
@@ -118,7 +118,7 @@ def unpackdx(fid, nrows):
         return tensMsh
 
     @classmethod
-    def readUBC(TensorMesh, fileName, meshdim=None):
+    def readUBC(TensorMesh, fileName, meshdim=None, folder=''):
```
lheagy: do we want `folder` or `directory` ? 
cc @rowanc1: do you know what is more common?

thast: Indeed directory does sound better.

discretize/MeshIO.py
```diff
@@ -118,51 +124,60 @@ def unpackdx(fid, nrows):
...
         if meshdim == None:
             # Read the file as line strings, remove lines with comment = !
-            msh = np.genfromtxt(fileName, delimiter='\n', dtype=np.str, comments='!', max_rows=1)
+            msh = np.genfromtxt(fname, delimiter='\n', dtype=np.str, comments='!', max_rows=1)
             # Fist line is the size of the model
             sizeM = np.array(msh.ravel()[0].split(), dtype=float)
             # Check if the mesh is a UBC 2D mesh
             if sizeM.shape[0] == 1:
-                Tnsmsh = TensorMesh._readUBC_2DMesh(fileName)
+                Tnsmsh = TensorMesh._readUBC_2DMesh(fname)
```
lheagy: to simplify this a bit (and reduce redundancy), why don't we just assign `meshdim = 2` here or `meshdim = 3` below, and then  check meshdim and call the appropriate `readUBC` method after that. 

Thoughts?

thast: I agree, this is a function I wrote in a previous PR but it was a redundant check between what the user expects and what the machine read.

thast:
Thanks @lheagy, can you merge it for me please? Does not seem I have the rights to do so 
("Only those with write access to this repository can merge pull requests.")

PR #86, opened 2018-01-19
ref/view

lheagy:
add the ability to mask streamlines if the amplitude is below a given threshold. 

image

Thoughts on the name of the keyword input: `mask_stream_below`?


thast:
Nice addition @lheagy !
For the name, how about something around `stream_threshold` or `stream_mask_threshold`?

lheagy:
Thanks @thast :) 

I like `stream_threshold`

PR #90, opened 2018-03-09
Fix writeVTK bug in python 3

thast:
fix to allow export models in VTK format in Python 3


thast:
Original bug:
image

jcapriot:
Were you getting this bug with the most recent version? Either way my original fix was just a simple quick hack, but I like this better.

thast:
@jcapriot : that is right, I was getting this error before your update. My apology, I did not realize you had fixed it. Thanks! I guess at this point it is more about which syntax we prefer. Also, it seems the warning is still here, both on master and my branch. I might as well look at it while I'm at it.

thast:
@jcapriot : Sorry but I don't get what you mean by " Although now the active model is the last key'd model and not the first."? I thought you might mean the ordering of the for loop but I have not notice a difference:
image
Thanks!

discretize/MeshIO.py
```diff
@@ -244,13 +244,13 @@ def writeVTK(mesh, fileName, models=None, directory=''):
...
+            for key in models:
                 # Convert numpy array
-                vtkDoubleArr = numpy_to_vtk(item[1], deep=1)
-                vtkDoubleArr.SetName(item[0])
+                vtkDoubleArr = numpy_to_vtk(models[key], deep=1)
+                vtkDoubleArr.SetName(key)
                 vtkObj.GetCellData().AddArray(vtkDoubleArr)
             # Set the active scalar
-            vtkObj.GetCellData().SetActiveScalars(list(models.keys())[0])
+            vtkObj.GetCellData().SetActiveScalars(key)
```
jcapriot: This line is outside the `for` loop, previously it was setting the first model in the dictionary as the active model. Here it is now setting the active model as the last value that `key` had in the for loop, (which was the last model in the dictionary)

jcapriot: It doesn't really effect anything though

lheagy:
@thast: would you mind updating the description to clarify how this builds upon #90? Thanks!

PR #94, opened 2018-05-02
Tree mesh acceleration

jcapriot:
At this point consider this branch as EXPERIMENTAL. There are many possible unsafe operations that could arise, so be careful (which will need to be enforced at a later time). The code is certainly not completely up to format standards, but at this point, just wanted to get a pull request going and allow anyone who wants to use it to test it out to find any bugs.

There are many changes to the TreeMesh implementation within this pull request, However it is mostly feature complete compared to the previous implementation with regards to the public members of the class.

It was mostly rewritten in a way that made the construction of the mesh and construction of the operators all done in c++/cython, which resulted in dramatic speedups. As a reference, the 97 nosetests on the tree in the previous implementation took 238.476s on my personal computer. This implementation took 20.281s.

The basic idea is that each object, (`node`, `edge`, etc.) is aware of the structure of the `TreeMesh` as each is a cpp class that contains references to other objects.

A few key differences:

- All tree construction is balanced (no need to call `tree.balance`, or pass `balance=True` to functions)

- `tree.refine` should only be called once (at this point) as it "finalizes" the tree. It might be good to add a flag to the tree initialization that would allow incremental additions (similar to `scipy.spatial.ConvexHull`) and then require a finalization operation to be done before any other operations.

- Interpolation is "lazy" 2nd order now for all `E`, `CC`, and `F` interpolation, meaning we triangulate the grid points to interpolate. Interpolating without triangulation on the TreeMesh is still a point of research. This is not as fast as it could be, but still faster than previously.

- Both `face` and `edge` operations are defined for 2D, (basically a re-ording of x-edges -> y-faces)

- Do not expect any ordering for the faces, edges, and nodes. They are whatever they have decided to be.

- Do not expect any of the private members of the class to have remained consistent between implementations.

- There are many other changes which should hopefully be apparent as you look through the code.

Big things that still need to be implemented

- Serialization, (do not expect to pickle this object and have it work, however with construction being much faster, this shouldn't slow down anyone too much while it is being worked on).

- `PlotImage` and `PlotSlice`.

There are also many areas that this code could be extended to handle different types of underlying meshs, as well as support for differing sizes of axes, but as I said before, the initial goal of this pull request is to mimic the behavior of the current implementation.

Other than that, I hope this speeds up the operations for those who need them.

Update
- Serialization should be implemented now (the `TreeMesh` is pickle-able)
- PlotImage and PlotSlice are also implemented.
- It also now support differing dimensions of the underlying `TensorMesh`
[...]

PR #96, opened 2018-07-12
Tree mesh accel merge

lheagy:
@jcapriot: I worked through the merge conflicts on #94 here. I wanted to do it as a pr so you could take a quick look.  Sorry it is a bit opaque b/c there are so many changes coming in from the 3D cyl mesh. I will tag where it is relevant to the tree mesh - the main item is that all of the `cellGradStencil`'s are properties rather than methods. I updated this and would appreciate a check from you. Feel free to give me a shout on slack if I can clarify anything!


discretize/tree_ext.pyx
```diff
@@ -1314,8 +1314,11 @@ cdef class _TreeMesh:
 
         return sp.csr_matrix((V, (I,J)), shape=(self.ntFz, self.nC))
 
+    @property
```
lheagy: @jcapriot: I added the property decorator and cache the cellGradxStencil (to prevent name mangling or double underscores, I appended `Mat` to the cached ones)

discretize/TreeMesh.py
```diff
@@ -166,11 +166,11 @@ def cellGradStencil(self):
         if getattr(self, '_cellGradStencil', None) is None:
 
             self._cellGradStencil = sp.vstack([
-                self._cellGradxStencil(), self._cellGradyStencil()
+                self._cellGradxStencil, self._cellGradyStencil
```
lheagy: now each component is a property (to be consistent with the other meshes and so that SimPEG can always grab them as properties individually)

discretize/tree_ext.pyx
lheagy: If construction is fast, we can also simply keep the decorator and not worry about caching the matrix. We just need to be able to treat it as a property in the SimPEG regularization

jcapriot: So this will work, only because when we use the Class it is as a parent to the actual class. This would not necessarily work on it's own in Cython because you cannot add members to the class on the fly. If you look at how I have to initialize all of the other possible cached operators you can see examples of it.

lheagy: ok, in this case, do you think it is better just to simply use the `@property` decorator and skip the caching?

discretize/TreeMesh.py
jcapriot: the caching will always work here because this code will always be executed in Python

discretize/tree_ext.pyx
lheagy: (I was looking at the `nodalGrad` for an example - but now I see that you create the cdef object on the `_TreeMesh` --> missed that on the first go)

jcapriot: I think it's just fine to cache them. Especially the stacked version in TreeMesh.py, as the stack operation is not inexpensive.

I think it sounds like you see it, it gets added as a cdef object at the beginning of the class, then is initialized to None in the __init__ method. Just have to think of cython cdef classes as a static thing, (like a class defined in c/c++).

Issue #100, opened 2018-07-20
Bug in travis early exits


lheagy:
The latest pypi release deployed earlier than anticipated (thanks @jcapriot for catching this). It wasn't a problem as the deployed version was what we intended to deploy - however, it should have been deployed only upon the merge to master or the tagged release...
```yaml
after_success:

  - bash <(curl -s https://codecov.io/bash)

  # early exit if not on a deplotment branch
  - if ! [ "$TRAVIS_BRANCH" = "$MASTER_BRANCH" -o "$TRAVIS_TAG" = "true" ]; then
      echo "Not deploying (because this is not a deployment branch)" ;
      exit 0 ;
    fi
  - if [ "$TRAVIS_PULL_REQUEST" = "true" ]; then
      echo "Not deploying (because this is a pull request)" ;
      exit 0 ;
    fi

```

PR #102, opened 2018-07-24
Include tree.pxd in source distribution

cgohlke:


jcapriot:
Never even thought to think of what was and wasn’t included in sdist. Definitely should be pulled in.

lheagy:
Excellent, thanks @cgohlke! I will wait until the travis test goes through and then pull this in

jcapriot:
I think the current fail has to deal with the current non-early exit bug in the Travis script

lheagy:
yeah, it seems to be related to #100

lheagy:
Thanks @cgohlke! This has been included in discretize 0.3.2 which is now available on pypi: https://pypi.org/project/discretize/ (windows wheels to follow shorty)

PR #104, opened 2018-08-06
plot_3d_slicer

prisae:
plot_3d_slicer
===

Add an interactive slicer for 3D volumes.  At the moment only implemented for tensor meshes.

plot3dslicer-2

Features
---
- Mouse wheel scroll while hovering over a subplot scrolls through the third axis (e.g., hovering over the xy-slice and scrolling your mouse wheel will go through the z-axis).
- The three subplots are synced, also for zooming and moving.
- The initial slices can be provided via the `xslice`, `yslice`, and `zslice` parameters (default is in the middle of the volume).
- Transparency values and ranges can be provided (a list of floats and tuples/lists of two values), e.g. to hide the seawater or to focus on an interesting part, e.g., `[0.3, [1, 4], [-np.infty, -10]]` to remove all values equal to 0.3, all values between 1 and 4, and all values smaller than -10. For interactive range selection set `transparency='slider'`.
- Takes `clim` and `pcolorOpts` as other mesh-plotting functions, which will be passed to `pcolormesh`.
- By default the horizontal axis is `x`, and the vertical axis is `y`; this can be flipped by setting `view='yx'`.

By default, the aspect ratio of the three subplots is set to `'auto'`. You can change this with the `aspect`
parameter, however, expect the unexpected by doing this. Most importantly, the three subplots won't be nicely aligned, and zooming might result in funny arrangements. Two parameters can be used in this respect:
- `aspect` takes `'auto'`, `'equal'`, or `num`. A list of two of them can be provided, in which case the first element is for the xy-slice, and the second element for the xz- and zy-slices. E.g., `aspect=['equal', 2]` sets the xy-slice to `equal`, and in the other two the vertical dimension is exaggerated by a factor of 2.
- The `plot_3d_slicer` is on a `subplot2grid`-grid, by default on a 3x3 grid, where 2x2 are used for the xy-slice, 2x1 for the xz-slice, and 1x2 for the zy-slice. You can provide a list of three integers via the `grid`-parameter, which stand for the number of grid-units occupied for the x-, y-, and z-dimension (default is `[2, 2, 1]`).

Usage
---
```
mesh.plot_3d_slicer(data)
```
It requires `%matplotlib notebook` in Jupyter. In regular IPython shells it should just work.

### DIFF
--- a/docs/examples/plot_slicer_demo.py
+++ b/docs/examples/plot_slicer_demo.py
@@ -22,7 +22,8 @@ from matplotlib.colors import SymLogNorm
 import sys
 
 if sys.version_info[0] < 3:
-    sys.exit("This example only runs on Python 3")
+    print("This example only runs on Python 3")
+    sys.exit(0)
 
 ###############################################################################
 # Download and load data


prisae:
I think it is good to go. There is more stuff that can be fit in or improved at a later stage, but as it stands I think it is stable and just working fine.

PR #105, opened 2018-08-06
ref/plotting

lheagy:
- allow kwargs for `line_color` and `line_width` in the plotgrid function
- helpful when plotting the mesh and model on a highly discretized mesh. e.g. 
commer_model

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -530,7 +530,7 @@ class TensorView(object):
 
     def plotGrid(
         self, ax=None, nodes=False, faces=False, centers=False, edges=False,
-        lines=True, showIt=False
+        lines=True, showIt=False, **kwargs
     ):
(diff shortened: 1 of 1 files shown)


discretize/View.py
```diff
@@ -618,6 +618,8 @@ def plotGrid(
                     marker="^", linestyle=""
                 )
 
+            line_color = kwargs.get('line_color', 'C0')
+            line_width = kwargs.get('line_width', 1.)
```
lheagy: I am not sure if these are the best keywords. In some ways, I think we might want to re-write this function as a class that has properties which are the line styles, etc. Thoughts?

craigmillernz: keeping the names the same as matplotlib would be useful 
so linewidth, not line_width
line_color would just be color

I also notice in Utils.PlotUtils.plot2Ddata that the default line width is 2, which is too fat (i've changed my version to 1).

PR #111, opened 2018-10-24
drop support for python 2.7

lheagy:
- add an f-string to the copyright in the __init__.py
- remove testing on python 2.7 on both travis and appveyor 
- include `requires_python >= 3.6` in the setup.py

- I suggest we move to version 0.4.0 with this pr


leouieda:
It's such a liberating feeling, isn't it?

banesullivan:
As an outsider who's been thinking about incorporating `discretize` with some of my projects, I would still need Python 2.7 compatibility. 

I've been toying with the idea of trashing all my [PVGeo](http://pvgeo.org) file IO for GIF Tensor/Tree Meshes and other `discretize`-friendly data formats. Then simply use the VTK interfaces already in `discretize` (I would also create VTK interfaces where they are missing in `discretize` and ensure the existing ones are good to go).

Still having Python 2.7 support would allow me to make [PVGeo](http://pvgeo.org) algorithms that call back to `discretize` and have a graphical user interface for `discretize` directly in ParaView. This would look similar to  [GMSGDataExchange/omf#27](https://github.com/GMSGDataExchange/omf/pull/27) which unfortunately won't be happening because the VTK interface in `omf` would have to be separated.

Me needing Python 2.7 support would only last until ParaView upgrades its Python to 3.6 which I believe is scheduled for Summer 2019.

prisae:
What about having `v0.3.3` on bug-fix-support till the end of 2019, but `v0.4.0` and future ones are Python 3 only? This would be the same as numpy/scipy will be anyway, just two months earlier. Or do you think you will need new features in `discretize` as well @banesullivan?

banesullivan:
@prisae, that would work fine! 

I could then set PVGeo's dependency as v0.3.3 and any VTK interfaces needed for `discretize` that are not currently implemented, I will add directly in PVGeo. Then once ParaView upgrades to Python 3, I would migrate those VTK object interfaces into discretize so classes like `CurvilinearMesh`, `CylMesh`, etc would have `toVTK()` methods like the `_toVTRObj` method in `MeshIO`.

lheagy:
Hi @banesullivan, thanks for chiming in! We can definitely continue to support in terms of bug fixes in the mean-time. 

P.S. I added you to the SimPEG org 🎉. I am thrilled that you are interested in interfacing to discretize - having accessible 3D viz is something that would have a very big impact!

setup.py
```diff
@@ -56,7 +56,7 @@ def configuration(parent_package='', top_path=None):
 
 setup(
     name="discretize",
-    version="0.3.3",
+    version="0.4.0",
     install_requires=[
         'numpy>=1.7',
         'scipy>=0.13',
```
prisae: How about taking the opportunity and crank up the scipy-requirement? Ideally at 1.0? But that might be too high. Anyway, 0.13 is a bit low, and leaving Python 2 behind might be a good point to adjust.

setup.py
```diff
@@ -78,5 +78,6 @@ def configuration(parent_package='', top_path=None):
     platforms=["Windows", "Linux", "Solaris", "Mac OS-X", "Unix"],
     use_2to3=False,
```
prisae: `use_2to3`: obsolete?

setup.py
lheagy: good call `>= 1`?

prisae: It might be a bit pushy. But it was released a year ago, and it was a big release (many changes), so it might make it easier...

banesullivan:
Would it be possible merge #114 to v0.3.x and deploy on PyPI before this bump to 0.4.0?

lheagy:
Yep, we can plan to do that @banesullivan. I am just not sure what is causing the upstream errors on python 2.7... it is failing on the pip install (I am sort-of hoping other groups are running into this and it will resolve itself, but it has been close to a week...)

lheagy:
Thanks @prisae - definitely forgotten :)

discretize/BaseMesh.py
```diff
@@ -717,13 +726,20 @@ def nFz(self):
...
 
     def r(self, x, xType='CC', outType='CC', format='V'):
-        """`r` is a quick reshape command that will do the best it
+        raise DeprecationWarning(
+            'This method will be depreciated in favor if `reshape`. Please use'
+            ' `reshape` instead.'
+        ):
+        return self.reshape(x, xType, outType, format)
+
+    def reshape(self, x, xType='CC', outType='CC', format='V'):
```
lheagy: I think we should also rethink input variable names here (e.g. `format` should not be used). What about 
```python
def reshape(self, x, x_loc='CC', out_loc='CC', out_format='V')
```

I think "loc" is a bit more descriptive (and even "location"). I am still not sure about "format". (It should also take "vec", "vector", "matrix" to let the user be a bit more flexible in the input). 

Thoughts? 

cc @rowanc1

prisae:
Talking about Python 3.5 over in https://github.com/simpeg/discretize/issues/219 it seems odd that this PR is still open. Could we merge this @lheagy ?

prisae:
Most of this was implemented in #227 instead.

PR #114, opened 2018-10-27
feat/vtk: Add VTK object interface

banesullivan:
### Description

I wanted to go ahead and get an MR going before the 0.4.0 release to see if there would be a benefit to having a VTK interface on the mesh classes. If this looks good, I hope to fill out all the `toVTK()` methods on a 0.3.x release.

Once merged, this would enable discretize to have a direct interface for VTK base software. Notably, I could then create file readers in [PVGeo](http://pvgeo.org) that would natively use discretize to read serialized discretize meshes, UBC mesh/model formats, etc.

### Progress

I implemented VTK conversion methods on the following classes:

- [x] TensorMesh
- [x] TreeMesh
- [x] CurvilinearMesh
- ~CylMesh~

TODO:
- [x] Add tests
- [x] Add VTK file I/O for each class
    - [x] Fix relative imports issue
    - [x] Migrate `TensorMeshIO.readVTK` to `vtkInterface`
- [x] Double check `vtkInterface._curvilinearMeshToVTK` as it looks strange in 3D when I plot it up in ParaView.
- [x] Fix issue for 2D Tree meshes

## Usage

See comments below


banesullivan:
@rowanc1, I went ahead and added a new `mixins` module and placed the `vtkInterface` class there which has static methods for VTK object conversion and file write outs. Note that this will only be available if VTK is available.

I left the `readVTK` function for `TensorMesh`s inside `MeshIO` because I was having issues with recursive imports when trying to relatively import into the `vtkInterface` module (see TODO statements).

## The API

Create your objects:
```py
import discretize
import numpy as np

sz = [16, 16, 16]
tens = discretize.TensorMesh(sz)
tree = discretize.TreeMesh(sz)

def refine(cell):
    if np.sqrt(((np.r_[cell.center]-0.5)**2).sum()) < 0.4:
        return 4
    return 3

tree.refine(refine)
curv = discretize.CurvilinearMesh(discretize.utils.exampleLrmGrid(sz, 'rotate'))
```

Now save them out:
```py
from discretize.mixins import vtkInterface as vtkInt

vtkInt.writeVTK('tens.vtr', tens)
vtkInt.writeVTK('curv.vtk', curv)
vtkInt.writeVTK('tree.vtu', tree)
```

### Backwards Compatability

Note that I left the `writeVTK` methods under `MeshIO` present which are only imported if VTK is available so that the Tensor and Tree meshes can still be saved with the old API. These are now wrappers for `vtkInterface.writeVTK`.

```py
tree.writeVTK(fileName='tree2.vtu')
tens.writeVTK(fileName='tens2.vtr')
```

banesullivan:
## The API

Create your objects:
```py
import discretize
import numpy as np

sz = [16, 16, 16]
tens = discretize.TensorMesh(sz)
tree = discretize.TreeMesh(sz)

def refine(cell):
    if np.sqrt(((np.r_[cell.center]-0.5)**2).sum()) < 0.4:
        return 4
    return 3

tree.refine(refine)
curv = discretize.CurvilinearMesh(discretize.utils.exampleLrmGrid(sz, 'rotate'))
```

Now save them out:
```py
tens.writeVTK('tens.vtr')
curv.writeVTK('curv.vtk')
tree.writeVTK('tree.vtu')
```

Note that all of this functionality is only imported if VTK is available. this is also consistent with the old API for Tensor and Tree meshes.

banesullivan:
CurvilinearMeshes look good to me!

```py
import discretize

sz = [16, 16, 16]
curv = discretize.CurvilinearMesh(discretize.utils.exampleLrmGrid(sz, 'rotate'))
vec = np.arange(curv.nC)
data = {'a': vec, 'b': vec + 1}

curv.writeVTK('curv.vts', models=data)
```

Then open that VTS file in ParaView and you get [**this rendering**](http://viewer.pvgeo.org/?fileURL=https://dl.dropbox.com/s/c5szhh38ttlh3fw/curvi.vtkjs?dl=0) which you can compare to `curv.plotGrid`:

```py
curv.plotGrid(lw='0.1')
```
*NOTE: I added `**kwargs` to `curv.plotGrid`.*

rowanc1:
That looks awesome. Thanks for posting a link! I really like how you just proxy the dropbox url!

banesullivan:
Almost done! I added the ability to specify the axes orientation of any given mesh. For an example of how to use this, let's say we have a 2D tree mesh that we want to define on the XZ plane. Lets first make our tree mesh object:

```py
import discretize

tree = discretize.TreeMesh( [16, 16])

def refine(cell):
    if np.sqrt(((np.r_[cell.center]-0.5)**2).sum()) < 0.4:
        return 4
    return 3

tree.refine(refine)

# Now save out the boring XY oriented tree mesh
tree.writeVTK('tree.vtu')
```

Now this createted an arbitrarily 2D tree mesh; arbitrary in the sense that the mesh is on a 2D plane but that plane could be rotated in any orientation in a cartesian space. To define what that orientation is, we can use the new `axis_*` properties:

```py
tree.axis_u = (1,0,0) 
tree.axis_v = (0,0,1) 
tree.axis_w = (0,1,0) 
tree._validate_orientation()
```

Now we have a `TreeMesh` explicitly defined on the XZ plane! This is quite useful for when we want to convert this to a VTK data object that must have its location in Cartesian space defined:

```py
tree.toVTK()
```

Or better yet, save it out for ParaView:

```py
tree.writeVTK('tree-rot.vtu')
```

This will look like the following where I have the first XY oriented mesh (in green) and the second XZ oriented mesh (in red):

tree-rot

discretize/BaseMesh.py
```diff
@@ -377,6 +377,46 @@ def copy(self):
         """
         return properties.copy(self)
 
+    axis_u = properties.Vector3(
+        'Vector orientation of u-direction',
+        default='X',
+        length=1
```
rowanc1: `1.0`? not sure if that will be a problem.

discretize/BaseMesh.py
```diff
@@ -377,6 +377,46 @@ def copy(self):
...
+    def cartesian(self):
+        """True if the axes are defined in the traditional <X,Y,Z> manner"""
+        if (    np.allclose(self.axis_u, (1, 0, 0)) and
+                np.allclose(self.axis_v, (0, 1, 0)) and
+                np.allclose(self.axis_w, (0, 0, 1)) ):
+            return True
+        return False
+
+    @property
+    def rotMtx(self):
```
rowanc1: I think spelling this out and moving to pep8? `rotation_matrix`?

discretize/BaseMesh.py
```diff
@@ -377,6 +377,46 @@ def copy(self):
...
+    def _validate_orientation(self):
+        """Check if axes are orthogonal"""
+        if not (np.abs(self.axis_u.dot(self.axis_v) < 1e-6) and
+                np.abs(self.axis_v.dot(self.axis_w) < 1e-6) and
+                np.abs(self.axis_w.dot(self.axis_u) < 1e-6)):
+            raise ValueError('axis_u, axis_v, and axis_w must be orthogonal')
+        return True
+
+    @property
+    def cartesian(self):
```
rowanc1: should this be `is_cartesian` or something indicating it is a bool? @lheagy thoughts?

discretize/mixins/vtkInterface.py
```diff
@@ -58,12 +60,13 @@ def __treeMeshToVTK(mesh, models=None):
...
         VTK_CELL_TYPE = vtk.VTK_VOXEL
         if ptsMat.shape[1] == 2:
-            # TODO: assumes any 2D tree meshes are on the XY Plane
             # Add Z values of 0.0 if 2D
             ptsMat = np.c_[ptsMat, np.zeros(ptsMat.shape[0])]
             VTK_CELL_TYPE = vtk.VTK_PIXEL
         if ptsMat.shape[1] != 3:
             raise RuntimeError('Points of the mesh are improperly defined.')
+        # Rotate the points to the cartesian system
+        ptsMat = np.dot(ptsMat, mesh.rotMtx)
```
rowanc1: Does this now rotate into 3D?

discretize/mixins/vtkInterface.py
```diff
@@ -141,6 +112,60 @@ def __createStructGrid(ptsMat, dims, models=None):
...
+        elif mesh.dim == 2:
+            nodes = np.c_[mesh.gridN, np.zeros((mesh.nN, 1))]
+        # Now garuntee nodes are correct
+        if nodes.shape != (mesh.nN, 3):
+            raise RuntimeError('Nodes of the grid are improperly defined.')
+        # Rotate the points based on the axis orientations
+        mesh._validate_orientation()
+        return np.dot(nodes, mesh.rotMtx)
+
+    def __tensorMeshToVTK(mesh, models=None):
```
rowanc1: should this be name mangled  with two `_` or just private?

rowanc1:
Does the `axis_u/v/w` update the grid properties? or just the vtk output? I would think it should change both?

rowanc1:
Also, is the mesh validated before it is passed to vtk objects? even just a `self.validate()` at the top?

discretize/mixins/vtkInterface.py
banesullivan: My reasoning for name mangling with two `_`s was that the `BaseMesh` class inherits the `vtkInterface` class and thus all of these functions would be available to any class inheriting the `BaseMesh` if they were private. By name mangling and providing a wrapper in the `toVTK` and `writeVTK` methods, we prevent users from accidentally calling the wrong conversion method.

If we made them private, then I'd be concerned about the ability to do this:
```py
>>> tens = discretize.TensorMesh([16, 16, 16])
>>> # not okay call:
>>> tens._treeMeshToVTK()
Exception: 'some sort of crazy error message'
```

However, there are a few name mangled methods in there that would make sense for any mesh type to have available:

- `__assignCellData`
- `__getRotatedNodes`

discretize/mixins/vtkInterface.py
banesullivan: I could be wrong about this... but this should rotate the points of a mesh from any orthogonal coordinate system to the Cartesian system given any `rotation_matrix`. 
In the prior lines, I ensure that the points are 3D before getting here. Thus you could have a 2D mesh where `axis_u` is defined as something other than `(1,0,0)` (similarly for `axis_v`) and the third dimension of points in that local frame will be set to zeros which will then get rotated to the Cartesian space.

discretize/BaseMesh.py
banesullivan: for the length? should it be an integer or a float? I'm not too familiar with `properties.Vector3`

banesullivan:
@rowanc1, on your comments:

> Does the axis_u/v/w update the grid properties? or just the vtk output? I would think it should change both?

What exactly do you mean?

> Also, is the mesh validated before it is passed to vtk objects? even just a self.validate() at the top?

Ah, I forgot to do this... I put it off because I wasn't sure if there was a validator that can take the models as well to make sure the models fit on the mesh's cell data

discretize/BaseMesh.py
lheagy: right now, we have the `_meshType` variable which sort-of achieves this (as long as it is not `CYL` it is cartesian). 

However, I do think it would be good to have meshes be aware of their coordinate type. What about a `coordinate_system` property? (e.g. `cartesian`, `cylindrical` or `spherical`). This would give us some more flexibility down-the road if we do add spherical to the mix

Then to see if a mesh is cartesian, it is `if mesh.coordinate_system == cartesian`

discretize/BaseMesh.py
```diff
@@ -373,6 +377,46 @@ def copy(self):
...
+    def _validate_orientation(self):
+        """Check if axes are orthogonal"""
+        if not (np.abs(self.axis_u.dot(self.axis_v) < 1e-6) and
+                np.abs(self.axis_v.dot(self.axis_w) < 1e-6) and
+                np.abs(self.axis_w.dot(self.axis_u) < 1e-6)):
+            raise ValueError('axis_u, axis_v, and axis_w must be orthogonal')
+        return True
+
+    @property
+    def is_cartesian(self):
```
lheagy: It is not actually clear to me if this will return False for a cyl mesh at the moment? Would it make sense for the meshes to have a `coordinate_system` property? it can take on the value `cartesian`, `cylindrical`, `spherical`

discretize/BaseMesh.py
banesullivan: I meant to achieve something else with this property. I want this property to be `True` if the coordinate reference frame (`axis_u`, `axis_v`, & `axis_w`) is *the typical Cartesian reference frame (<1,0,0>, <0,1,0>, <0,0,1>)* rather than a rotated Cartesian system. 

In hindsight, I gave this a poor name. Suggestions on a clearer name? Perhaps `refrence_is_not_rotated` or something similar? Or maybe I shouldn't have it as a propery because this is only needed for when I convert `TensorMesh`es to VTK data objects. If the `TensorMesh` is defined on a rotated frame I have to use a `vtkStructuredGrid` instead of the typical `vtkRectilinearGrid`.

banesullivan: Went ahead and did this!

discretize/BaseMesh.py
banesullivan: Done in fb5caa7

discretize/BaseMesh.py
lheagy: is there a reference for UVW mapping you could add to the docstrings? It may not be a familiar concept to everyone

discretize/mixins/vtkModule.py
```diff
@@ -0,0 +1,490 @@
...
+    The two above code snippets produce a ``vtkRectilinearGrid`` and a
+    ``vtkStructuredGrid`` respecitvely. To demonstarte the difference, we have
+    plotted the two datasets next to eachother where the first mesh is in green
+    and its data axes are parrallel to the traditional cartesian reference frame.
+    The second, rotated mesh is shown in red and its data axii are rotated from
+    the traditional cartesian refence frame as specified by the ``axis_u``,
+    ``axis_v``, and ``axis_w`` properties.
+
+    .. image:: ../images/vtk-rotated-example.png
+
```
lheagy: Thanks for the excellent docs @banesullivan!

banesullivan:
@lheagy, thanks for looking over all of this! I think it is good to go with only a few minor features that might best fit in a new MR.

  

Tasks left for another MR:

- [ ] Overhaul the rotation to handle any origin
- [ ] Create ‘gridN_rotated’ property
- [ ] Do a validation check before VTK conversion (see #122)

discretize/BaseMesh.py
banesullivan: I don’t have any good resources on hand for the UVW axes conventions but I’ll search around and see what I can pull together and put in the docs.

lheagy:
🎉 @banesullivan, whenever you get the chance, please add yourself to [AUTHORS.rst](https://github.com/simpeg/discretize/blob/master/AUTHORS.rst)

PR #115, opened 2018-10-28
fix/py27: Fix the Travis build for Python 2.7

banesullivan:
There was some weirdness happening with the pip installed from conda on py2.7. To mitigate this, I just set the build to use the native py27 pip for all dependencies. If Python version is 3.x then it uses conda to manage dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,15 @@ before_install:
   - conda update --yes conda
 
 install:
-  - conda install --quiet --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib pillow cython ipython;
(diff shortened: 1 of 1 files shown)


banesullivan:
Looks green except for the `tests/docs` task. Not sure why the warning is triggering a failure:
```
(line  112) 
Warning, treated as error:
/home/travis/build/simpeg/discretize/docs/index.rst:112: WARNING: broken link: http://slack.simpeg.xyz/ (504 Server Error: Connection Timed Out for url: http://slack.simpeg.xyz/)
FAIL
```

lheagy:
@rowanc1: would you mind updating the godaddy pointer for `slack.simpeg.xyz` to "https://simpeginvite.herokuapp.com/" (I think we only had http before and that might be causing a delay in the re-direct...). Thanks!

PR #130, opened 2018-12-28
Add vtki support to make using VTK data objects more Pythonic

banesullivan:
I recently added a ton of new features to [`vtki`](https://github.com/akaszynski/vtki) (the VTK interface Python package) that help make using just about any VTK data objects more intuitive/Pythonic. If `vtki` is available in your Python environment then these changes make using VTK data objects way easier. Here's a screenshot of what this currently looks like in a Jupyter notebook (creates static renderings but can also be interactive in a separate pop-up window). 

Also check out [this notebook](https://github.com/OpenGeoVis/PVGeo-Examples/blob/master/3.0%20-%20Using%20the%20vtkInterface%20Package.ipynb) to see more ways to use [`PVGeo`](https://github.com/OpenGeoVis/PVGeo), `discretize`, and [`vtki`](https://github.com/akaszynski/vtki). Below is a screenshot of a simple use case:

ezgif com-video-to-gif

screen shot 2018-12-28 at 2 27 21 pm

Some sample code to do it yourself!

```py
import discretize
import numpy as np

# Create a simple TensorMesh
h1 = np.linspace(.1, .5, 3)
h2 = np.linspace(.1, .5, 5)
h3 = np.linspace(.1, .5, 3)
mesh = discretize.TensorMesh([h1, h2, h3])

# Convert to VTK object and use vtki to render it
mesh.toVTK().plot(notebook=False)
```

### DIFF
--- a/discretize/mixins/vtkModule.py
+++ b/discretize/mixins/vtkModule.py
@@ -52,6 +52,7 @@ from vtk import vtkXMLUnstructuredGridWriter
 from vtk import vtkXMLStructuredGridWriter
 from vtk import vtkXMLRectilinearGridReader
 
+import warnings
 
 
 def assignCellData(vtkDS, models=None):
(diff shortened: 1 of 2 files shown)


lheagy:
So the pr is failing because the government is shut-down, (haha?)... 

image

It is a broken link in the docs which is completely unrelated to this work: https://www.ctcms.nist.gov/fipy/examples/cahnHilliard/generated/examples.cahnHilliard.mesh2DCoupled.html

@banesullivan: I will try to find a suitable replacement link and patch it.

banesullivan:
> So the pr is failing because the government is shut-down, (haha?)...

🤣 

> @banesullivan: I will try to find a suitable replacement link and patch it.

Sounds good! Thanks, @lheagy!!

PR #134, opened 2019-01-29
Documentation cleanup

### LINKED ISSUE #126: numpy-styled docs
Similar to simpeg/simpeg#703, we should consider using numpy-styled docs for discretize. They are much cleaner to read inline, which is very relevant for when people pull up docs in Jupyter. 

https://numpydoc.readthedocs.io/en/latest/format.html

### LINKED ISSUE #127: Organization of docs
# The problem

While looking at the style of the docs (#126), the organization should also be fleshed out. (we have part of the API documented in the "Tutorial"...)
image. 

I am also debating if we want to try and separate the API docs from the User Documentation. In some ways, I think this might be good. We could describe the classes and provide a brief overview of its properties and methods and link each to the appropriate API page (this is approximately what numpy does). 

# A suggestion

I see the items in bold as headings in the table of contents side bar (see for example the [JOSS docs](https://joss.readthedocs.io/en/latest/reviewer_guidelines.html))
**Overview**
- Why discretize?
- Getting started (which should be improved to include common windows problems) 
- Examples

**User Documentation**
- Finite volume (includes variable locations and terminology)
- Tensor Mesh
- Cylindrical Mesh
- Tree Mesh
- Curvilinear Mesh
- Differential Operators
- Inner Products
- Utilities
- Visualization
- VTK Module

**API**
- API

# Feedback?

I would appreciate thoughts here! (@leouieda, @prisae: you have both written beautiful docs - any suggestions here??)

### LINKED ISSUE #128: organization of base classes
Users shouldn't really be using the BaseMesh, yet we are importing it in the `__init__.py`. I think we should remove this. Any concerns with this?

lheagy:
# Improvements
## Organization of base classes
- move base classes to a `base` directory (closes #128)
 
## Docs
- new docs can be previewed at: ~~https://discretize.appspot.com~~ http://discretize.simpeg.xyz/en/repo-cleanup/
- use napoleon + numpy-style docs to compile the docs (closes #126) 
    - convert existing docstrings to numpy-styled docs 
- separate the API documentation from user documentation (closes #127)

image

## Testing
- travis cleanup (previously it was confusing which version of python was being used. We used the python 3.6 image on travis but then downloaded the latest conda - which is python 3.7): now each test suite is clearly labeled
- use pytest for testing instead of nose

# Todos:
~~- [ ] depreciate rtd in favour of google app engine site https://discretize.appspot.com (cc @rowanc1: can you help here :) )~~
- [x] fix the auto-deploy of the docs
- [x] fix coverage 
- [x] create an issue outlining the content to be developed in the "User Guide" (see #149) 
- [x] create a documentation guide (separate pr) for how we document classes, methods, functions and class attributes (e.g. https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) (see #150)
- [x] pr on SimPEG to ensure it is up-to-date with the changes in the base-class don't cause upstream problems (see simpeg/simpeg#776)

# Thanks:
- @leouieda : for your beautiful repo-setup and docs to follow :)


leouieda:
@lheagy you've been busy lately 🙂 Any reason for using app engine over GitHub pages?

rowanc1:
I think just due to how we were doing four years ago. Can GitHub pages be under your own domain?

lheagy:
@leouieda: yeah, as @rowanc1 said, it was what we were familiar with. I would be open to GH pages. Can you have multiple subdomains with GH pages? (e.g. discretize.simpeg.xyz, docs.simpeg.xyz, ...? or would we need to reorganize to simpeg.xyz/docs/, simpeg.xyz/discretize/?)

leouieda:
You can configure subdomains by adding `CNAME` files. But I wasn't suggesting the switch, just curious if there was any advantage to using app engine :)

I don't think there is much advantage to using ghpages if you have a nice setup working. Do you have to pay thing for app engine? If so, then that might be an incentive to move to github. The scripts in https://github.com/fatiando/continuous-integration can deploy to Github from TravisCI but there is also the [Doctr](https://github.com/drdoctr/doctr) which looks really good and I should probably be using that now that I think about it.

lheagy:
Thanks @leouieda, I don't think there is much of an advantage of one over the other. If you have dynamic content, google will support that, but for docs that isn't really necessary. We have billing set up, but the site is lightweight enough that it doesn't cost us anything at the moment.

Thanks for the links - I have been borrowing a few ideas from your continuous integration repo - I really like the setup!

lheagy:
Instead of deploying to google, this branch now builds on rtd (which simplifies the deploy!) http://discretize.simpeg.xyz/en/repo-cleanup/

dccowan:
Comments
=======

General Structure
-------------------

Really good. But I would like to see use add a theory section so we describe what the finite volume approach and the associated operators are doing. Suggested sections:
- Getting started
- User guide (I would like this to be tutorial-based examples with explanation)
- API
- Theory
- Getting connected

lheagy:
Hi @dccowan, many thanks for the review! I appreciate the idea of a "Theory" section, but I am wondering if this would be better suited for a geosci module. I think this might simplify the content that we maintain here, and keep the separation that geosci is meant to be more of the "textbook" resource with all of the theory. 

For example, much of SimPEG for EM theory can point to http://em.geosci.xyz. Based on the course material that we assembled for the course that @dougoldenburg and @sgkang delivered in Argentina, I think we have enough to start http://inversion.geosci.xyz (or http://computation.geosci.xyz), and we could put the theory for finite volume here. 

This would keep the docs more scoped to describe how someone uses the software, a few details on how things are implemented, and the remaining reference material could live in geosci. 

What do you think of this separation?

PR #135, opened 2019-02-12
Pure Python 3D Viz Example

banesullivan:
Here are some updates to the `vtkModule` docs that reflect recent development in `vtki` and a new example for 3D viz. Check out the example code and its image below:

vtki_laguna_del_maule

### For the reviewer:

- [x] I'm not sure if this example is getting executed/tested during the `make html`.
- [x] How do I use the screenshot I created under `docs/images/vtki_laguna_del_maule.png` as the example icon?


banesullivan:
I updated the example so that it now shows a thumbnail in the gallery - it's ready to merge!

Issue #142, opened 2019-03-20
Improve `print(mesh)`


prisae:
I think the current print-utility of meshes was made with small meshes in mind. It sort of explodes for big, stretched meshes:

![screenshot](https://user-images.githubusercontent.com/8020943/54692397-3ebda580-4b25-11e9-989e-1c93f1a68bb2.png)

I think something along these lines would be preferable (maybe as a rendered table in notebooks):

    discretize.TensorMesh: 128 x 128 x 128 cells; 2,097,152 in total
    
        dir          min/max N            min/max h
        -----------------------------------------------------
         x:      -4176.3 / 4176.3;     25.0 / 143.0
         y:      -5902.6 / 5902.6;     50.0 / 160.8
         z:      -7151.7 / 7151.7;     20.0 / 355.8

prisae:
Related, to some extent, to #21.

banesullivan:
I could help implement a  `_repr_html_` method like what I did for `vtki`:

<img width="730" alt="Screen Shot 2019-03-20 at 10 20 54 AM" src="https://user-images.githubusercontent.com/22067021/54701087-e2915c00-4af9-11e9-8eb9-483b92455cea.png">

prisae:
I think that would be great! But I think @lheagy and @rowanc1 and probably others of SimPEG have to chime in with their opinion, as it is their brainchild.

A few points:
- What happens if you are not in a notebook? Is there a plain text fallback?
- I think minimum cell-width in each distance would be useful.
- Maybe format the bounds as `.1f` instead of `.3e`

banesullivan:
The `_repr_html_` method is only ever used in an IPython environment. If you want a plain text representation to fall back on, then it can be implemented in the `__repr__` method:

<img width="504" alt="Screen Shot 2019-03-20 at 12 39 09 PM" src="https://user-images.githubusercontent.com/22067021/54710390-2e013580-4b0d-11e9-90be-eb1c037d5370.png">

prisae:
It would also go nicely along with the `versions`. 
![Selection_002](https://user-images.githubusercontent.com/8020943/54710461-03809e00-4b48-11e9-9cb4-ecb5beb861c1.png)

prisae:
That is interesting. I never figured out, how to distinguish between a Notebook kernel and an IPython/Qt kernel. If you do the same in an IPython terminal, what will happen?

prisae:
[...]

PR #143, opened 2019-03-20
Add representation methods for TensorMesh

banesullivan:
# Add representation methods for `TensorMesh`

The current `TensorMesh`-string method is somewhat outdated, particularly for big meshes. This PR adds html and non-html representations which should be more generally applicable, for small and big `TensorMesh`'s.

Based on work by @banesullivan on `vtki` and @prisae on the `printversion`-tool.

Selection_001


prisae:
Great, thanks for this quick draft. Could you put a screenshot of a more complicated model, along the lines of #142 ?
```
discretize.TensorMesh(
    [[(25, 59, -1.03), (25, 10), (25, 59, 1.03)],
     [(50, 59, -1.02), (50, 10), (50, 59, 1.02)],
     [(20, 59, -1.05), (20, 10), (20, 59, 1.05)]],
    x0='CCC')
```

Also, my points from there hold:
> - I think minimum cell-width in each distance would be useful.

So besides X/Y/Z Bounds also min(X/Y/Z) (which is simply min(mesh.hx), max(mesh.hx)).

Also, maybe split the number of cells. Maybe leave `Dimensionality` out, but then `Number of cells: 10 x 20 x 30`, so the dimensionality can be deduced from it.

Here again the 'sketch' from #142:
```
discretize.TensorMesh: 128 x 128 x 128 cells; 2,097,152 in total

    dir          min/max N            min/max h
    -----------------------------------------------------
     x:      -4176.3 / 4176.3;     25.0 / 143.0
     y:      -5902.6 / 5902.6;     50.0 / 160.8
     z:      -7151.7 / 7151.7;     20.0 / 355.8
```

prisae:
That is a really good starting point. I'll see if I get around soonish to polish it

banesullivan:
```py
discretize.TensorMesh(
    [[(25, 59, -1.03), (25, 10), (25, 59, 1.03)],
     [(50, 59, -1.02), (50, 10), (50, 59, 1.02)],
     [(20, 59, -1.05), (20, 10), (20, 59, 1.05)]],
    x0='CCC')
```
Screen Shot 2019-03-20 at 2 09 59 PM


We could do some restructuring of the table layout that's currently there to accommodate that information.

lheagy:
Very cool! I think this would be a great improvement to discretize. I don't think the` print(mesh)` has seen much use, so a cleaner version that is useful for you will likely be useful to others 

(also I had not seen a "draft" pr before - that is nice feature)

prisae:
First go at it, inputs welcome.

@banesullivan, I simplified some stuff, because I didn't quite understand it. Maybe I simplified to much, can you have a look.

@fourndo, I don't know if that also works for TreeMeshes, could you have a go? Similar for CylMeshes @lheagy.

Selection_001
Selection_002

Just saw that I miss the tiles in the non-html version.

prisae:
I improved the text-version:

screenshot

prisae:
@banesullivan, you put it into `discretize/BaseMesh.py`. I moved it now to `discretize/TensorMesh.py`, and replaced there the old `__str__` function within the `TensorMesh`-class that served for printing until now.

This means it is only for TensorMeshes and not other meshes, right now. I think this is the least friction, so it should be straight forward to merge. Similar steps should be taken to replace the `__str__`-functions in other meshes.

prisae:
screenshot_2

prisae:
The failures are related to Python 2.7. But it is an easy one to avoid, will do it.

prisae:
:smile: Sorry @banesullivan, messed that up with assigning. Wanted to assign you as reviewer. But I think as you originally created it it doesn't take you as reviewer...

banesullivan:
This looks really good @prisae! +1

## Afterthoughts

> This means it is only for TensorMeshes and not other meshes, right now. I think this is the least friction, so it should be straight forward to merge.

Why can't this be generalized in the `BaseMesh` class? Should there eventually be a generalized version of this?

A few thoughts arise on other attributes to include:

- What about things like `volume`/`area`? - I'm often most concerned with the general size of a mesh. Would this be something worth including?
- Should there be a function that can print meta info about a scalar dictionary? `discretize` doesn't exactly enforce any types for data attributes to a mesh but it might be nice to have a function that can take a dictionary of arrays and print out descriptive statistics on those arrays - perhaps `min`, `max`, `size`, `mean`, `std`. I'm thinking this might look like [pandas' `.describe` function](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.describe.html). The function could be associated with the mesh and help decipher if arrays in that dictionary lie on the nodes, cells, or faces of a mesh based on their size.
    - This might spark a different discussion on whether or not `discretize` should be able to handle `pandas.DataFrame`s and `Series`

prisae:
Thanks @banesullivan for the feedback!

> Why can't this be generalized in the BaseMesh class? Should there eventually be a generalized version of this?

Well. There are a couple of thoughs:
1. The currently `__str__`implementation is for each mesh separate, so it is the easiest to just improve them. And I guess there is a reason why they are in the different meshes, and not in `BaseMesh`.
2. I never worked with cylindrical or OctTree meshes, so I wouldn't know what is important there. So again, I am going the easy path here.
3. I think different meshes have very distinct characteristics. For instance cylindrical meshes and tensor meshes are quite different, so I think it is actually best to keep this as a per-mesh basis.

So, in summary. Maybe it could be combined into a more general thing into `BaseMesh`. However, this would need input from various other people, and it might take some effort and time to get it done (and might get buried under other things and never done). So I preferred to do what I can and get it included. But besides that, I doubt if it makes sense to generalize it too much so it fits into `BaseMesh`. (Pinging here @lheagy, @rowanc1, @sgkang, @fourndo for other opinions.)

Regarding your first point in the list: I am generally interested about the extent of a mesh, and the number of cells in a mesh. The first point to see if I cover my model and my sources/receivers, the second one to know if my memory will be able to handle the calculation. So I personally am not bothered about `volume` nor `area`. But this might be different for other folks.

I think it would be good if you create a separate issue with your second point in the list.

lheagy:
This is looking great! A couple questions / thoughts
- the notation `min(N)` `max(N)` isn't immediately evident to me, the N is for node? Is there space to spell out `node` instead of the `N` or does that get cluttered?
- It might be nice to include the origin as a column?
- Similar to @prisae's comment: I haven't come across many cases where the volume or area of the entire mesh is a particularly insightful quantity 
- For cylindrical meshes: they aren't particularly different than Tensor Meshes (the Cylindrical Mesh class inherits the TensorMesh class). For the information shown in these print-outs, I think the print-out for the Cyl mesh could be identical to the Tensor
- The OcTree Mesh is quite different, here are a few parameters I think might be useful to show. @jcapriot, @fourndo might have a few other ideas to share.  
    - number of total cells
    - number of levels of refinement
    - smallest cell size along each dimension
    - largest cell size along each dimension
    - origin

discretize/TensorMesh.py
```diff
@@ -598,56 +598,65 @@ class TensorMesh(
...
-            outStr += '\n  nCz: {0:d}'.format(self.nCz)
-            outStr += printH(self.hx, outStr='\n   hx:')
-            outStr += printH(self.hy, outStr='\n   hy:')
-            outStr += printH(self.hz, outStr='\n   hz:')
-
-        return outStr
+    def _get_attrs(self):
+        """An internal helper for the representation methods"""
+        attrs = []
+        dims = ['x', 'y', 'z']
```
lheagy: this could be a property on the mesh - it would make it easier to inherit (then on the cyl mesh for example, all of this code could be re-used, but `dims = ['r', 'theta', 'z']`

prisae:
>     * the notation `min(N)` `max(N)` isn't immediately evident to me, the N is for node? Is there space to spell out `node` instead of the `N` or does that get cluttered?

Yes, `N` was meant for node. I think `min(node)` and `max(node)` would fit without any issues. Think about it, I thought simply `min`, `max` might be sufficient and even clearer? It is simply min/max of the dimension.
 
>     * It might be nice to include the origin as a column?

Will do, good input!

prisae:
> * It might be nice to include the origin as a column?

Thinking about it: What do you mean with origin? that is basically what `min(N)` did so far, right?

prisae:
Have a look at the new screenshot. I changed `min(N)` to `origin`, and `max(N)` to `extent`. Is it clearer like that, or any other suggestions?

Also, I converted `_get_attrs` into a property `attributes`, and instead of a tuple a made a dict, so that the content is properly named. Also, I added the property `dim_names`, this is what would change then in cylmesh, correct? Please have a look @lheagy and let me know.

prisae:
The alternative to a `attributes`-dict would be to create directly parameters along the line of `x0`:
- `self.x1 = (extent x, extent y, extent z)`
- `self.h0 = (min(hx), min(hy), min(hz))`
- `self.h1 = (max(hx), max(hy), max(hz))`
- `self.f0 = (min factor x, min factor y, min factor z)`
- `self.f1 = (max factor x, max factor y, max factor z)`

Opinions?

discretize/TensorMesh.py
```diff
@@ -327,3 +332,45 @@ def cellBoundaryInd(self):
...
+    @property
+    def dim_names(self):
+        """Names of the dimensions.
+
+        Different for different meshes.
+        """
+        return ['x', 'y', 'z'][:self.dim]
+
+    @property
+    def attributes(self):
```
lheagy: I appreciate that this is handy, but I am a bit hesitant at this point to add another property to the Tensor Mesh that contains much of the same information as is already available through other properties (e.g. the list in the docs is pretty extensive already... http://discretize.simpeg.xyz/en/master/api/generated/discretize.TensorMesh.html#module-discretize.TensorMesh) 

What do you think of performing these actions in the print statement?

prisae: @lheagy, you confuse me here... :smile:

I made this a property because in your previous review (https://github.com/simpeg/discretize/pull/143#pullrequestreview-224688874) you said "this could be a property on the mesh - it would make it easier to inherit (then on the cyl mesh for example, all of this code could be re-used". But I am happy to go back to what it was.

discretize/TensorMesh.py
```diff
@@ -60,56 +60,60 @@ class TensorMesh(
...
+
+        # Get attributes and put into table.
+        attrs = self.attributes
+        for i in range(self.dim):
+            name = self.dim_names[i]
+            iattr = attrs[name]
+            fmt += "<tr>"  # Start row
+            fmt += "<td>{}</td>".format(name)
+            fmt += "<td>{}</td>".format(iattr['nC'])
+            for p in ['origin', 'extent', 'h_min', 'h_max', 'max_fact']:
```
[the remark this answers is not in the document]
prisae: Yes, let's agree on naming before I make any further changes: At the beginning I had `min(N)`/`max(N)`. You thought that is not very intuitive, which I agree, and we should add the origin. But the origin is the same as the minimum. So I changed `min(N)`/`max(N)` to `origin`/`extent`.

I think the possibilities are:

- `min`, `start`, `origin`, `x0`
- `max`, `end`, `extent`

Having in mind that there is also `min(h)`/`max(h)` and we should avoid confusion.

Another idea might be to have a double-line header:
```
              ==  TensorMesh: 2,097,152 cells  ==

                         mesh extent            cell width   str fact
  dir    nC           min           max       min       max       max
 --------------------------------------------------------------------
   x    128          0.00      8,352.67     25.00    143.00      1.03
   y    128     -5,902.58      5,902.58     50.00    160.83      1.02
   z    128    -14,303.35          0.00     20.00    355.79      1.05

```

discretize/TensorMesh.py
lheagy: Sorry for the confusion! In the previous review, I was only referring to the `dimension_names`, and it was really only after working on #134 that I got a sense of how over-loaded the attributes are.

If you prefer to leave this as a property, I think it is okay, as long as we make it a private variable e.g. `_attributes`

discretize/TensorMesh.py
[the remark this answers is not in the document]
lheagy: I quite like the double-line header example you have listed last. It displays a lot of useful info, and I think is fairly clear to interpret.

prisae:
I made a double header line in both, plain text and html. Also, the `attributes` is private again, `_repr_attributes`. I used quite some additional vertical space (more lines) in the html-implementation, but I think it is clearer and easier to read this way.

PR #141, opened 2019-03-19
Feat/refine tree

fourndo:
- [x] Add functionality for TreeMesh creation
- [x] General bug fix on refine function

PR #144, opened 2019-03-22
Initial mesh tools branch for TensorMeshes

prisae:
# Add more meshing tools to create fancy meshes

## Motivation
We probably all have our own little routines to create more complex meshes. I think it would be great to incorporate some of these into discretize, so not everyone has to do the same again.

## Status so far
Here, I included my most used routines:
- `get_domain`: A routine that calculates minimum cell width and calculation extent based on multiples of the skin depth. The user can restrict the domain to a minimum domain, and give a range for the possible min-cell width. Good for on-the-fly mesh creations.
- `get_stretched_h`: Get the input (together with `get_domain`) for a `TensorMesh` for a certain domain size, a certain number of cells, a minimum cell width, and an optional domain without stretching. The routine finds the suitable stretching factor.

## Warning
- This is an initial draft. I tried to document as good as possible, so it should be easy to incorporate other ideas and stuff from other routines. Please add your own stuff!
- I added tests for 100% coverage. However, in pytest, so they probably will fail hilariously. I thought I won't bother checking if they are good for unittest as long as it is in an early stage.
- I also added an initial notebook, didn't check, will probably fail too.

### DIFF
--- a/docs/examples/plot_meshutils.py
+++ b/docs/examples/plot_meshutils.py
@@ -1,0 +1,59 @@
+"""
+``meshutils`` to create complicated meshes
+==========================================
+
+The examples demonstrate some of the tools in `utils.meshutils`.
+"""
+
+# %matplotlib notebook
+import discretize
+import numpy as np
+import matplotlib.pyplot as plt
+import sys
+
+if sys.version_info[0] < 3:
+    print("This example only runs on Python 3")
+    sys.exit(0)
+
+###############################################################################
+# 1. Get stretched grid with one stretching factor for various restrictions
+# -------------------------------------------------------------------------
+#
+# - Source at 500 => Domain is centered around this point
+# - Frequency 10 Hz, Resistivity 1 Ohm m
+# - 2^5 cells
+h_min, domain = discretize.utils.meshutils.get_domain(x0=500, freq=10, rho=1)
+
+hx = discretize.utils.meshutils.get_stretched_h(h_min, domain, nx=2**5, x0=500)
+grid = discretize.TensorMesh([hx, 1], x0=[domain[0], 0])
+grid.plotGrid()
+
+###############################################################################
+# 1.b As before, but with following additional restrictions
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# - My model goes from -800 to 3500, so I want this extent at least.
+# - I want the minimum cell width to be within 5-100 meters, no smaller, no
+#   wider.
+# - From 300 to 800 meter I want regular spacing with min cell width.
+
+h_min, domain = discretize.utils.meshutils.get_domain(
+    x0=500, freq=10, rho=1, limits=[-800, 3500], min_width=[5, 100])
+
+hx = discretize.utils.meshutils.get_stretched_h(
+        h_min, domain, nx=2**5, x0=300, x1=800)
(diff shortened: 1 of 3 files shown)


prisae:
Files changed:
- `discretize/utils/discretize/meshutils.py` => Routines
- `docs/exapmles/plot_meshutils.py` => Example
- `tests/base/test_utils.py` => Tests

Let me know if they would go better somewhere else.

prisae:
Here is the notebook, which I also added as an example:

https://github.com/prisae/tmp-share/blob/master/TensorMesh-utilities.ipynb

docs/examples/plot_meshutils.py
```diff
@@ -0,0 +1,59 @@
+"""
```
rowanc1: Thanks for adding the example!

thast:
Just to mention it, @sgkang and @lheagy had a utils to create 1D mesh for MT in their Leading Edge MT tutorial. Maybe worth a look too: https://github.com/simpeg/tle-magnetotelluric_inversion/blob/49b038b013a40283a32ea18ea9144bf707656873/tleMT/MT1D.py#L72

prisae:
Thanks @thast, exactly things like that could be added. I personally won't add anything more, the ones I added are the ones I use in my current project, but I invite everyone to add more tools. Although this could also be done at a later stage, of course.

I meant this PR also to discuss the topic in general. Is `meshutils` the right place for it? Or should we start a dedicated file just for such routines? Any input/idea/suggestion around the topic of mesh generation is welcomed here. 

My meshing routine is only good in some circumstances. For instance, if you have a fixed number of cells you want to use (like I use only powers of two, 2^x, currently). If you need large padding, like in my case, I set the air-padding to something like 40*skindepth. But obviously, for other applications these settings won't suit.

fourndo:
That's great @prisae . I worked on something similar, mostly targeting for 3D TreeMesh on #141 
Called it ```meshBuilderXYZ``` right now.
Maybe we should see if we can mash-up the two ...

prisae:
@fourndo that would be great and exactly what I am looking for. To combine our many approaches and create a few widely applicable helper routines.

prisae:
Had a look at your function. So the main difference is:
- Your function uses a fixed stretching factor, and finds the required number of cells.
- My function uses a fixed number of cells, and tries to find the stretching factor. 

On top of that, you have the whole Tree-Mesh stuff.

An underlying design question is what it returns. You return a 3D mesh. In my case, it just returns the cell widths, you still have to call TensorMesh with it. The downside of my approach is that more lines of codes are needed when you use it, but the upside is that it keeps it much more general, as it can be used for 1D, 2D, and 3D.
```
hx = discretize.utils.meshutils.get_stretched_h(hx_min, xdomain, nx, x0, x1)
hy = discretize.utils.meshutils.get_stretched_h(hy_min, ydomain, ny, y0, y1)
hz = discretize.utils.meshutils.get_stretched_h(hz_min, zdomain, nz, z0, z1)
mesh = discretize.TensorMesh([hx, hy, hz], x0=[xdomain[0], ydomain[0], zdomain[0]])
```

I think if we want to combine them the first step would be to split up yours into a similar scheme, so it only returns cell widths, with which you can call TensorMesh. If this is something you don't want to consider, then we keep them better as separate options.

prisae:
I am closing this. It is sort of outdated, I think everybody moved on and uses some functionalities etc.

It would still be great if we have some automated meshing functionalities, but I think we would have to start fresh.

PR #145, opened 2019-03-25
Add `fig`-parameter to `plot_3d_slicer`

prisae:
# Add `fig`-parameter to `plot_3d_slicer`

Small change to provide more flexibility for `plot_3d_slicer`. It now takes an optional `fig`-parameter (100% backwards compatible), in which one can provide an existing figure handle. The figure is cleared at every call, but no new figure is created. This can give some more flexibility, for instance in the use together with widgets.

Initiated upon an idea by @fourndo, who uses it for a Mag Tutorial:
Peek 2019-03-25 20-24
where he used the figure handle to replace the YZ-plot with a data plot, and wrap it into a widget.

Code example: Now you can call
```
fig = plt.figure()
mesh.plot_3d_slicer(model, fig=fig)
```
where `fig` is an existing figure. And then you can do more funky stuff with your figure handle. It is sort of a convenience addition, as the same would be possible with:
```
fig = plt.figure()
tracker = discretize.View.Slicer(mesh, model, **kwargs)
```

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -737,7 +737,7 @@ class TensorView(object):
     def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
                        vType='CC', view='real', axis='xy', transparent=None,
                        clim=None, aspect='auto', grid=[2, 2, 1],
-                       pcolorOpts=None):
+                       pcolorOpts=None, fig=None):
         """Plot slices of a 3D volume, interactively (scroll wheel).
 
         If called from a notebook, make sure to set
@@ -759,9 +759,15 @@ class TensorView(object):
 
           `fig.get_children()[0].get_children()`.
 
+        One can also provide an existing figure instance, which can be useful
+        for interactive widgets in Notebooks.
+
         """
         # Initiate figure
-        fig = plt.figure()
+        do_show = False
+        if fig is None:
+            do_not_show = True
+            fig = plt.figure()
 
         # Populate figure
         tracker = Slicer(self, v, xslice, yslice, zslice, vType, view, axis,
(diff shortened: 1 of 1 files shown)


prisae:
@fourndo, if it works, maybe you can link your notebook as an example.

There are still things that are potentially possible, but probably not that straight forward. E.g., if you scroll/zoom/pin now your figure, and then change something in the widget, the scroll/zoom/pin status will all be reset to the defaults. It would definitely be possible to get this info and pass it back in, but not sure how straight forward it is.

It could be easier by calling straight the `Slicer`, and add a functionality to just change the data. But then when I tried the Slicer directly with the widget it didn't like it.

That whole thing might be a rabbit hole...

So this is sort of an easy and quick fix. You can use it with a widget, it does not create a new figure, but it resets the whole figure nevertheless. Everything else might take much more time (or might require the help of a matplotlib-guru).

lheagy:
🎉 Excellent, we can release this as discretize 0.3.11. @prisae: would you mind adding a couple more details to the main thread above? e.g. just a little bit of context on why the widgets didn't work before and do now. We will use your commentary as the release notes. Thanks!! 🚀

prisae:
@lheagy thanks for the fast reply. I probably removed the 'Draft'-label to early. I would keep that PR open until @fourndo has his example finished. Because maybe there will come more. And on top, when he finishes, we have a nice figure :smile: 

(What is ready to be included from my point of view, on the other hand, is #143.)

prisae:
@lheagy, OK, this one is good to go after discussion with @fourndo. Added the required additional information.

lheagy:
Excellent, thanks @prisae and @fourndo!

PR #146, opened 2019-03-26
Correct '--inline' to '--inplace'.

prisae:
Typo in the docs.


jcapriot:
I’ve always thought that you should not recommend installing discretize in this manner, as setting PYTHONPATH can get cumbersomb when running different python environments.

Installing it in this manner is really only useful if you are actively developing discretize

lheagy:
does `pip install -e .` work? if so that takes care of the pythonpath for you (I just am not sure if that deals with the c bindings correctly?)

prisae:
Yeah, maybe that should be stated better. Give me a second.

prisae:
What do you think to the change, @jcapriot, @lheagy ?

jcapriot:
I think if you're installing from source (and not intending to actively change your install of discretize and just use it). You should just run a "python setup.py install" Which will build the compiled libraries, and then install it to the active python's site_packages directory.

prisae:
Not sure, there were some issues, @lheagy might know more. Feel free to edit my PR though @jcapriot!

I only intended to fix a typo. It states `python setup.py build_ext --inplace` in `discretize/utils/interputils.py` and in `discretize/__init__.py`, but `--inline` in docs/content/installing.rst`, which is clearly wrong.

prisae:
(I actually don't use `PYTHONPATH`, but have a symlink for `discretize` to the directory where I need it. Whether that is better or worse I don't know...)

jcapriot:
There shouldn't be anything wrong with running a setup.py install. The only issues I've seen with people are from ones who didn't quite understand the ramifications of setting PYTHONPATH, or the order that python looks for packages, which is first the current working directory, then PYTHONPATH, then site_packages.

Therefore if you startup python from the base directory of discretize, it will prioritize running the local version in the current working directory. Which might cause tests to fail because the extensions don't get built locally with a "setup.py install".

jcapriot:
@prisae It's a perfectly acceptable thing to do for someone who is actively modifying and changing discretize itself.

lheagy:
:wave: @jcapriot: would you be willing to add some language to explain the python path a bit to the docs? The text from your last comment is helpful.

lheagy:
I will get this in right after #134 (that should fix the travis fail)

PR #153, opened 2019-04-24
Module names

lheagy:
- use lower case names for modules (TensorMesh.py --> tensor_mesh.py)
- note that this doesn't change the import structure of discretize. 
  ```
  from discretize import TensorMesh
  ```
  still works as expected
- to run the TreeMesh, you will need to re-install
  ```
   python setup.py build_ext --inplace
  ```

Todos:
- fix the slicer example: the `shelve` is not unpacking correctly because of this change. @prisae, would you be willing to help update this example?


prisae:
I tried to run the laguna_del_maule notebook, but due to your changes this doesn't work either, I think.

If you can do it, all you have to do is to put these lines after the inversion:
```
with shelve.open('./laguna_del_maule-result') as db:
    db['mesh'] = mesh
    db['Lpout'] = Lpout
```
This will generate you a new archive, which you can upload then instead of the existing.

prisae:
The error I get, in case that helps you:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-cc408050a04a> in <module>
      2 import shutil
      3 import tarfile
----> 4 import SimPEG.PF as PF
      5 from SimPEG import (
      6     Maps, Regularization, Optimization, DataMisfit,

~/anaconda3/lib/python3.7/site-packages/SimPEG/__init__.py in <module>
      3 
      4 import discretize as Mesh
----> 5 from discretize import Tests
      6 
      7 from . import Maps

ImportError: cannot import name 'Tests' from 'discretize' (/home/dtr/Codes/discretize/discretize/__init__.py)
```

prisae:
Closed as it was superseeded by #227

PR #152, opened 2019-04-22
PyVista updates & deploy docs to GitHub Pages

banesullivan:
### Changes

- Update `discretize` to work with PyVista (previously `vtki`)
- Enable the PyVista 3D visualization examples to be run when making the docs on CI services
- Switch the documentation hosting service to GitHub Pages from Read The Docs
- New `InterfaceOMF` mixin for converting `discretize` meshes to Open Mining Format (OMF) objects
- `pep8` refactoring of `mixins`
- Drop Windows testing on Python 3.5

### Todo

1. **SQUASH AND MERGE** into master (these changes need to be squashed)
2. Make sure docs are deploying to `en/master/` directory under https://github.com/simpeg/discretize-docs
3. Switch `discretize.simpeg.xyz` domain CNAME to `simpeg.github.io`

PR #154, opened 2019-04-24
include the vtki demo in the gallery

lheagy:
- examples for the gallery of the docs are now in the `examples` directory rather than the `docs/examples` directory
- moved the vtki demo so that it shows up in the gallery!


banesullivan:
Hey, thanks @lheagy! 

This might co flirt with #152

lheagy:
Thanks for pointing that out @banesullivan! You are on-top of things :) 

I will close this in favour of #152

Issue #155, opened 2019-05-01
Soft dependency for matplotlib


prisae:
At the moment `matplotlib` is a hard dependency (in the `setup.py`-file). Is there anything that speaks against making it a soft dependency just like `vtki`?

=> it is "only" used for the plotting of meshes, which is sort of a relatively small (but, agree, important) part of the whole `discretize` scope.
=> It would help to install discretize on minimal conda-environments for running models on a server, without having to install matplotlib.

rowanc1:
Yeah, totally agree with that. Moving this into a soft dependency would be good.

lheagy:
Completely agree!

lheagy:
closed in #156

PR #156, opened 2019-05-01
Make matplotlib a soft dependency

prisae:
Addresses issue #155 .

Here is a first go at it. I only included it in `View.py` and `Test.py`, and removed `matplotlib` from `setup.py`; the `@requires`-decorator is in `utils/codeutils.py`. So I expect it will fail hilariously on every testing platform.

Any opinions @lheagy, @rowanc1?

I don't think I can do that all on myself. I would need help with the tests (unittest, need worked with those), and probably with everything non-pure-python in `discretize` (@jcapriot ?).

Ready for discussion!


prisae:
I started off with the idea from `pysal` (https://github.com/pysal/pysal/blob/master/pysal/lib/common.py), I think a decorator is ideal for this use case. However, I'd prefer if all imports are at the beginning of each file, with a `matplotlib=False` if it fails, not like `pysal` with their `simport`.

lheagy:
:+1: This looks like a pretty clean way to do this! cc @rowanc1, @jcapriot: do you have any other thoughts on how to make this dependency soft in a clean fashion?

One other option might be to do all viz imports through `view.py`: down the road this could allow us to plug in different plotting back-ends, but for a first step, we can keep it simple.

discretize/Tests.py
```diff
@@ -356,8 +364,8 @@ def l2norm(x):
         print(sadness[np.random.randint(len(sadness))]+'\n')
 
 
+    @requires({'matplotlib': matplotlib})
```
rowanc1: I had no idea you could `@wrap` if statements. Cool.

discretize/View.py
```diff
@@ -1011,6 +1018,7 @@ def plotImage(self, *args, **kwargs):
         return self._plotCylTensorMesh('plotImage', *args, **kwargs)
 
 
+@requires({'matplotlib': matplotlib})
```
rowanc1: Thoughts on this going on the class not the plotting methods?

prisae: I thought about this too. My reason to put it on the lower level was in case that the class is also needed for other stuff which I didn't want to break, and I didn't want to check everything... But I would be all in to move it up to the class level.

discretize/Tests.py
lheagy: hmm, this seems to error on me. Is there something I need to have installed or updated? (That is what is causing the travis fail too...) 

Maybe we just put these lines of code in a function?

discretize/View.py
lheagy: At this point I don't have strong opinions either way, so I am fine leaving this as is, and later on we could move it up if that makes sense? 

This also could be nice down the line if we want to support other plotting backends (e.g. plotly)

prisae:
Probably not the cleanest solution, but if it works, it works ;-)

prisae:
@lindsey, the tests passed, so I removed the 'Draft'-tag of the PR.

lheagy:
Thanks @prisae!

prisae:
> One last minor thing, lets add `matplotlib` to `requirements_dev.txt`.

Done!

PR #157, opened 2019-05-02
Mesh tutorials

dccowan:
This PR is meant to create a section for tutorials. We started with tutorials for generating
- tensor meshes
- cylindrical meshes
- tree meshes
There were also a couple of bug fixes. One in the surface-based refinement for QuadTree meshes. And another in the 'plot_3d_slicer' method.


discretize/View.py
```diff
@@ -778,7 +778,10 @@ def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
         fig.canvas.mpl_connect('scroll_event', tracker.onscroll)
 
         # Show figure
-        plt.show()
+        if showIt:
+            plt.show()
+
+        return fig
```
prisae: We might want to think about the `return fig`. It has the `fig` parameter in the call, so you can provide an exiting figure, in which case you don't want/need it to be returned. Is it always required?

But anyhow, as long as it is streamlined with `discretize` then I'm fine with it.

discretize/View.py
```diff
@@ -737,7 +737,7 @@ def plotGrid(
     def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
                        vType='CC', view='real', axis='xy', transparent=None,
                        clim=None, aspect='auto', grid=[2, 2, 1],
-                       pcolorOpts=None, fig=None):
+                       pcolorOpts=None, fig=None, showIt=False):
```
prisae: I probably would prefer `showIt=True` as default. Otherwise I have to provide `showIt=True` every time I call the slicer and I want to see it right away? Or am I missing something?

prisae:
Thanks @dcowan. I just made two comments in the slicer class. I think you're changes will make it indeed more aligned with other functions in `discretize` (I wouldn't call it bug though, just adding features :wink: )

I never fully understood the logic behind the `showIt`-functionality (that is why I didn't implemented it in the first place probably), but I assume it has to do with examples/notebooks.

Anyhow, my two comments are just thoughts/ideas. I am fine with anything that is streamlined with the rest of discretize.

discretize/View.py
lheagy: I think the original motivation for the `showIt` was really for testing (cc @rowanc1). We didn't have the `agg` backend set on travis, so showing the plots made a mess. At this point, I think we can leave it out, or at the very least set the default as `True`

discretize/View.py
lheagy: The main motivation for the return here would be if someone didn't provide it as an input, but instead let the figure handle be created by the function. Then if you want to add labels and such after-the-fact, you can access the figure handle

rowanc1:
The `showIt` was for the examples and testing, and before a lot of the niceties we have now. So I am not sure how relevant it is? If we can get rid of it, that would be great!

discretize/View.py
rowanc1: Yeah, I think this is used a lot in modifying figures after the fact. Returning the thing you created seems like good practice.

docs/demo_examples/plot_3D_cyl.rst
```diff
@@ -0,0 +1,74 @@
+.. note::
```
lheagy: is this auto-generated? if so, then we don't need to include it in the repository

docs/demo_examples/plot_cahn_hilliard.py
```diff
@@ -0,0 +1,116 @@
...
+    # (phi_ - phi)/dt = A*d2fdphi2*phi_ + A*( - d2fdphi2*phi + dfdphi - L*phi_)
+    # phi_ - phi = dt*A*d2fdphi2*phi_ + dt*A*(- d2fdphi2*phi + dfdphi - L*phi_)
+    # phi_ - dt*A*d2fdphi2 * phi_ =  dt*A*(- d2fdphi2*phi + dfdphi - L*phi_) + phi
+    # (I - dt*A*d2fdphi2) * phi_ =  dt*A*(- d2fdphi2*phi + dfdphi - L*phi_) + phi
+    # (I - dt*A*d2fdphi2) * phi_ =  dt*A*dfdphi - dt*A*d2fdphi2*phi - dt*A*L*phi_ + phi
+    # (dt*A*d2fdphi2 - I) * phi_ =  dt*A*d2fdphi2*phi + dt*A*L*phi_ - phi - dt*A*dfdphi
+    # (dt*A*d2fdphi2 - I - dt*A*L) * phi_ =  (dt*A*d2fdphi2 - I)*phi - dt*A*dfdphi
+
+    h = [(0.25, n)]
+    M = discretize.TensorMesh([h, h])
```
lheagy: I realize this is probably copy-over, but lets try and stick with `mesh` instead of `M` for mesh objects in the examples

lheagy:
👋 Hi @dccowan, I see you closed this pr - do you have another working branch in progress?

Issue #158, opened 2019-05-14
Broken Example: plot_3d_slicer


dccowan:
The example *plot_slicer_demo.py* breaks. This is related to *View.plot_3d_slicer*. In line 66 of the example, *fig = plt.gcf()*, we don't obtain an object with the axes from the previous plot. So *fig.get_children()* is empty.

Why aren't we using a *showIt* option and returning the figure like many of the other plotting tools?

dccowan:
We should also control the figure size. Some of the text is overlapping and it looks messy.

lheagy:
closed with #159. Feel free to re-open @dccowan if there are still issues here

PR #159, opened 2019-05-14
Fix plot_3d_slicer bug

dccowan:
This addressed issue #158 . In this solution we include a *showIt=False* as an argument for the method. We also output the figure. The size of the figure has been custom set to avoid overlapping text in the example.


discretize/View.py
```diff
@@ -737,7 +737,7 @@ def plotGrid(
     def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
                        vType='CC', view='real', axis='xy', transparent=None,
                        clim=None, aspect='auto', grid=[2, 2, 1],
-                       pcolorOpts=None, fig=None):
+                       pcolorOpts=None, showIt=False, fig=None):
```
lheagy: The `showIt` should not be necessary here. Please see the discussion we had in the previous pr: https://github.com/simpeg/discretize/pull/157#discussion_r280660843

dccowan: Sure. That is fine.

dccowan: It looks like the fix for now is to return figure instead of automatically plotting

discretize/View.py
```diff
@@ -777,8 +777,7 @@ def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
         # Connect figure to scrolling
         fig.canvas.mpl_connect('scroll_event', tracker.onscroll)
 
-        # Show figure
-        plt.show()
+        return fig
```
lheagy: cc @prisae: just a small functionality update here I wanted to bring you in on. It seems like there are some use-cases where having access to the figure handle outside would be handy if it is generated inside this function. Does this work for you?

examples/plot_slicer_demo.py
```diff
@@ -50,6 +50,7 @@
 # 1.1 Default options
 # ^^^^^^^^^^^^^^^^^^^
 
+fig = plt.figure(figsize=(12, 12))
```
lheagy: This line is not needed. It creates an additional figure in the example. If you needed to use a custom figure size, you could pass it to the `plot_3d_slicer` e.g. `mesh.plot_3d_slicer(Lpout, fig=fig)`, but for the demo, I think it is simplest if we stick with the defaults and let `mesh.plot_3d_slicer` generate the figure
```suggestion
```

image

discretize/View.py
prisae: Yes, no problem, I don't have any hard feelings with regards to how this is handled. I did not overly adjust the `plot_3d_slicer` from my usage/style to SimPEG-style. I think the idea  was we just bring it in, and then adjust it where necessary, so perfectly fine by me.

examples/plot_slicer_demo.py
lheagy: :wave: @dccowan: do you have any concerns if I commit the change suggested above?

dccowan: I decided to set the figure size because some of the text was overlapping. I removed the line a regenerated and for all intents and purposes it's actually not bad.

dccowan:
This seems to fail on a link check. Not sure which file. But I don't believe there are any other problems.

docs/conf.py
```diff
@@ -112,6 +112,7 @@
     'http://docs.scipy.org/doc/numpy/reference/generated/numpy.array.html#numpy.array',
     'http://dx.doi.org/10.1016/j.cageo.2015.09.015',
     'http://slack.simpeg.xyz',
+    'http://docs.simpeg.xyz/content/examples/04-grav/plot_laguna_del_maule_inversion.html#sphx-glr-content-examples-04-grav-plot-laguna-del-maule-inversion-py',
```
lheagy:
```suggestion
```

PR #163, opened 2019-05-15
Change model array shape check for VTK mixin

### LINKED ISSUE #162: to_vtk is too lenient with model array sizes
@thast pointed out to me that there is an issue when passing non-vectorized data arrays to the `to_vtk()` method.

```py
import pyvista as pv
import numpy as np
import discretize

mesh = discretize.TensorMesh.readUBC('Mesh_model/mesh3d.msh')

model3d = np.load('Mesh_model/model.npy')
print(model3d.shape, model3d.size)
```
((10000, 28), 280000)

Then this:

```py
mesh.to_vtk({'model3d': model3d})
```

works fine but as soon as you try to use that PyVista mesh, the VTK backend will segfault.

The reason you are able to create the PyVista object has to do with this check:

https://github.com/simpeg/discretize/blob/c16e34c93fd0b0b2848751ddde5cccbaadb5b82f/discretize/mixins/vtkModule.py#L88-L89

Should we change this to check the first axes of the model array? This way users could pass arrays containing vector data (`nc` by 3):

```py
...
if mod.shape[0] != nc: 
     raise RuntimeError(...
```

banesullivan:
Resolve #162

### DIFF
--- a/discretize/mixins/vtkModule.py
+++ b/discretize/mixins/vtkModule.py
@@ -85,8 +85,8 @@ def assign_cell_data(vtkDS, models=None):
     if models is not None:
         for name, mod in models.items():
             # Convert numpy array
-            if mod.size != nc:
-                raise RuntimeError('Number of model cells ({}) does not match number of mesh cells ({}).'.format(mod.size, nc))
+            if mod.shape[0] != nc:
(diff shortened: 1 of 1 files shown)


thast:
I can review when this is ready.

banesullivan:
Thanks, @thast - It's ready when the CIs give the green check!

thast:
That looks good! Thanks @banesullivan. I will merge now and do a release tomorrow (maybe the plot_slice bug fix PR can go in as well...)

PR #166, opened 2019-05-16
Add `(x,y,z)lim` to plot_3d_slicer

### LINKED ISSUE #165: Add ranges to plot_3d_slicer
`plotImage` and `plotSlice` have the `range_x` and `range_y` parameters. We should include them (plus `range_z`) in the 3D slicer too.

I can think of two possibilities:
- Restrict `mesh` and `v` in the `__init__`-call. This is the easiest implementation. But then you cannot move the figure to data outside of the range in the interactive plot.
- Set the limits in the initial plot. But then we have to take care how the interactive zooming etc works, and the home-button to go back to the initial view.

prisae:
Add `(x,y,z)lim` to limit the axis. It works with the interactive tools and the home-button will reset to the provided limits. Resolves #165.

Peek 2019-05-21 08-25

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -736,7 +736,7 @@ class TensorView(object):
 
     def plot_3d_slicer(self, v, xslice=None, yslice=None, zslice=None,
                        vType='CC', view='real', axis='xy', transparent=None,
-                       clim=None, range_x=None, range_y=None, range_z=None,
+                       clim=None, xlim=None, ylim=None, zlim=None,
                        aspect='auto', grid=[2, 2, 1], pcolorOpts=None,
                        fig=None):
         """Plot slices of a 3D volume, interactively (scroll wheel).
@@ -774,13 +774,14 @@ class TensorView(object):
         # Populate figure
         tracker = Slicer(
             self, v, xslice, yslice, zslice, vType, view, axis, transparent,
-            clim, range_x, range_y, range_z, aspect, grid, pcolorOpts
+            clim, xlim, ylim, zlim, aspect, grid, pcolorOpts
         )
 
         # Connect figure to scrolling
         fig.canvas.mpl_connect('scroll_event', tracker.onscroll)
 
-        return fig
+        # Show figure
+        plt.show()
 
 
 class CylView(object):
@@ -1218,7 +1219,7 @@ class Slicer(object):
     clim : None or list of [min, max]
         For pcolormesh (vmin, vmax).
 
-    range_x, range_y, range_z : None or list of [min, max]
+    xlim, ylim, zlim : None or list of [min, max]
         Axis limits.
 
     aspect : 'auto', 'equal', or num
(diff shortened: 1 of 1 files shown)


banesullivan:
Would `xlim`, `ylim`, `zlim` be better than `rang_*`. This is more consistent with `matplotlib`: https://matplotlib.org/api/_as_gen/matplotlib.pyplot.xlim.html

prisae:
Potentially. But this is in agreement with the existing tools (`plotSlice`, `plotImage`). I don't want to start using different parameters. Your suggestion could be used when the whole plotting utilities of discretize would get a work-over...

banesullivan:
> Your suggestion could be used when the whole plotting utilities of discretize would get a work-over...

IMHO, I'd go ahead and use more `Matplotlib` friendly syntax now rather than waiting for a refactor down the road.

Another option which we did for some plotting keywords in PyVista was allow aliasing... You could have `range_*` be a declared keyword argument with a default value of `None` and check other passed arguments in the `**kwargs` for things like `xlim` when `if `range_x is None:`

We did this for a lot of stuff in PyVista - off the top of my head, the color bar limit argument can be `clim` or `rng`

prisae:
I would be all in for that. Still, I would not make the change in this PR (which is about bringing the slicer to same functionalities as other tools), but create another PR changing the `range_*` to `*lim`, if the main `discretize` developers have an interest in that (pinging @rowanc1, @lheagy, @grosenkj, @fourndo, @jcapriot ).

lheagy:
There is a bit of a subtle difference between the behaviour of `xlim` and the `range_x` in the plotSlice. `xlim` in matplotlib still draws the whole figure and then restricts the limits on the axes that are displayed. The `range` values actually pull out a piece of the model and plot only that, so if you were to resize the xlim externally, that portion of the plot is non-exsistent. For example

image

(at least for vector plots, there is an inconsistency at the moment if only scalars are plotted) 

So I opted to use a different name because the behaviour is a bit different. But I am totally open to discussion here!

prisae:
In this case I should actually call it `xlim`, `ylim`, and `zlim`, as I use that behaviour in the slicer, I do not cut it up, as you can see in the GIF. What do you think?

prisae:
Or, alternatively, would you prefer if I cut the volume also in the slicer as you do in plotSlice?

lheagy:
That sounds good @prisae! Sticking with matplotlib notation where it mimics that behaviour is simplest

prisae:
Done. From my point of view, this is good to go.

lheagy:
:wave: @thast, would you be willing to do the merge and release here? Steps:
- [ ] merge pr + delete branch
- [ ] pull the latest master to your local machine
- [ ] `git status` (make sure there are no lingering changes from your local work, if there are please stash them or start a new branch)
- [ ] `coda patch` updates the version number to 0.4.4
- [ ] `git push` to push these changes to GitHub 
- [ ] generate release notes based on the pr description (e.g. https://github.com/simpeg/discretize/releases/tag/0.4.3) 
    - the new release should autodeploy to pypi and appveyor when the testing finishes
- [ ] celebrate on slack 🎉

prisae:
Good list @lheagy! Is that already noted somewhere? Otherwise it might be a good list to add to a MAINTENANCE.rst-file for future reference, instead of this PR...

thast:
Good idea @prisae !

Issue #167, opened 2019-05-20
Restrict Sphinx < 2.0


prisae:
The docs are also affected by this problem:

https://github.com/numpy/numpydoc/issues/215

![screenshot](https://user-images.githubusercontent.com/8020943/58012788-8d6fc780-7af5-11e9-8485-2f9b79357fc3.png)

So I suggest to restrict Sphinx to < 2.0. I will make a PR in this regard in a sec.

prisae:
Was implemented by #168.

PR #168, opened 2019-05-20
Restrict Sphinx<2.0

prisae:
Addresses #167.


lheagy:
Many Thanks @prisae 🎉

prisae:
No bother. We should keep an eye out when this will be resolved. But it seems to affect A LOT of docs, so hopefully it won't take too long.

PR #174, opened 2019-05-31
Update TensorMesh-OMF interface

banesullivan:
At the moment, only Tensor meshes are supported by OMF.

OMFv2 should bring more support for Curvilinear and Tree meshes.

In the meantime, here are updates to the TensorMesh-OMF interface to make going to/from OMF/discretize more fluid.


banesullivan:
Hey, I totally forgot about this PR. I'll add some tests and make sure all is working as expected.

lheagy:
Thanks @banesullivan :) Let me know if there is anything I can help with.

banesullivan:
@lheagy: All should be finished once the CI's give the check

PR #177, opened 2019-06-25
Feat/tree functions

micmitch:
The addition of several functions from @jcapriot to make it easier to work with octree meshes. Several of the functions are used to integrate line source terms for FDEM. Other improvements include updating documentation and improving plotGrid to visualize octree meshes and models.


discretize/TreeMesh.py
```diff
@@ -310,21 +389,98 @@ def faceDivz(self):
...
+            Av = self._getEdgeIntMat(locs, zerosOutside, locType[1])
+        elif locType in ['Fx', 'Fy', 'Fz']:
+            Av = self._getFaceIntMat(locs, zerosOutside, locType[1])
+        elif locType in ['CC']:
+            Av = self._getCellIntMat(locs, zerosOutside)
+        return Av
+
     @property
     def permuteCC(self):
+        """Permutation matrix re-ordering of cells sorted by x, then y, then z"""
```
lheagy: Thanks for the doc-strings :)

jcapriot:
I don't have any more local things left on this so I'm good to pull this in

lheagy:
It seems that only python 2.7 is giving issues here. I am inclined to just remove testing and indicate that we no longer explicitly support python 2.7. Are there any concerns with this?

PR #182, opened 2019-07-22
serialization of basemesh 

lheagy:
update instantiation of basemesh so meshes can be formed from their serialized state, e.g.
```
mesh = discretize.BaseMesh.deserialize(mesh_json)
```
will have the value of `_n` stored, thus, `n` does not need to be provided on instantiation

### DIFF
--- a/discretize/TreeMesh.py
+++ b/discretize/TreeMesh.py
@@ -84,6 +84,9 @@
 #      |           |         ^
 #      |___________|         |___> x
 #      0    e3     1
+
+import properties
+
 from .base import BaseTensorMesh
 from .InnerProducts import InnerProducts
 from .MeshIO import TreeMeshIO
@@ -101,7 +104,12 @@ class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
     _meshType = 'TREE'
 
     #inheriting stuff from BaseTensorMesh that isn't defined in _QuadTree
-    def __init__(self, h, x0=None, **kwargs):
+    def __init__(self, h=None, x0=None, **kwargs):
+        if 'h' in kwargs.keys():
+            h = kwargs.pop('h')
+        if 'x0' in kwargs.keys():
+            x0 = kwargs.pop('x0')
+        # print(h, x0)
         BaseTensorMesh.__init__(self, h, x0)#TODO:, **kwargs) # pass the kwargs for copy/paste
 
         nx = len(self.h[0])
@@ -230,6 +238,10 @@ class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
         full_tbl += "</table>\n"
 
         return full_tbl
+
+    @properties.validator('x0')
+    def _x0_validator(self, change):
+        self._set_x0(change['value'])
 
     @property
     def vntF(self):
--- a/discretize/TreeMesh.py
+++ b/discretize/TreeMesh.py
@@ -121,6 +121,11 @@ class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
         # Now can initialize cpp tree parent
         _TreeMesh.__init__(self, self.h, self.x0)
 
+        if 'cell_levels' in kwargs.keys() and 'cell_indexes' in kwargs.keys():
+            inds = kwargs.pop('cell_indexes')
+            levels = kwargs.pop('cell_levels')
+            self.__setstate__((inds, levels))
+
     def __repr__(self):
         """Plain text representation."""
         mesh_name = '{0!s}TreeMesh'.format(('Oc' if self.dim==3 else 'Quad'))
@@ -644,8 +649,17 @@ class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
             plt.show()
         return tuple(out)
 
-    def save(self, *args, **kwargs):
-        raise NotImplementedError()
+    def serialize(self):
+        serial = BaseTensorMesh.serialize(self)
+        inds, levels = self.__getstate__()
+        serial['cell_indexes'] = inds
+        serial['cell_levels'] = levels
+        return serial
+
+    @classmethod
+    def deserialize(cls, serial):
+        mesh = cls(**serial)
+        return mesh
 
     def load(self, *args, **kwargs):
         raise NotImplementedError()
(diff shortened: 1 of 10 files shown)


tests/tree/test_tree.py
```diff
@@ -111,6 +111,18 @@ def test_faceDiv(self):
 
         self.assertEqual((M.faceDiv - T.permuteCC*T.faceDiv*T.permuteF.T).nnz, 0)
 
+    def test_serialization(self):
+        hx, hy = np.r_[1., 2, 3, 4], np.r_[5., 6, 7, 8]
+        mesh1 = discretize.TreeMesh([hx, hy], levels=2, x0=np.r_[-1, -1])
```
jcapriot: it might actually be useful to do something with the TreeMesh, i.e. `mesh1.refine(2)`

lheagy:
Got it sorted @jcapriot: It was a subtle error with the behaviour of `np.int32` related to numpy/numpy#2543

lheagy:
:wave: @jcapriot: whenever you have a chance, would you mind taking another look and if you think it is good to go then approve the pr? or if there are still changes you see fit, please let me know. Thanks!

jcapriot:
@lheagy It looks good to me. Can you look at the TreeMesh serialize/deserialize too?

discretize/TreeMesh.py
```diff
@@ -632,8 +649,17 @@ def plotSlice(
             plt.show()
         return tuple(out)
 
-    def save(self, *args, **kwargs):
```
banesullivan: Could this method stay and allow a user to pass a file name which will serialize any mesh to a file? For example:

```py
mesh = some_discretize_mesh_of_any_type
mesh.save('my_awesome_mesh.mesh')
```

jcapriot: It was removed because It was originally overriding it here to explicitly say that save wouldn't work on the TreeMesh, now it behaves like the other meshes

banesullivan: The `msh` and `mesh` extensions are greatly overused across viz libraries and other mesh analysis software. I think it would be best to step away from the `msh`/`mesh` convention I've seen used with discretize and towards something more distinct.

I think this PR would be a great place to standardize how meshes are saved - perhaps have the default extension to this `save` method be something like: `.dmesh`, `.discretize`, `.simpeg`

banesullivan: > It was removed because It was originally overriding it here to explicitly say that save wouldn't work on the TreeMesh, now it behaves like the other meshes

Ah whoops - I wasn't looking closely enough  - I see now that `save` is indeed implemented here for all meshes

https://github.com/simpeg/discretize/blob/7c0af69f2c67ffe92e7759f93a9eb98df6d7fb85/discretize/base/base_mesh.py#L395

banesullivan: My note about the file extension remains - instead of `.json`, would something more distinct be appropriate?

lheagy: Thanks @banesullivan: yes, the save / load should be working for all mesh-types at this point in discretize. 

I am inclined to leave the file extension as `json` as we can then take advantage of viewers and and tools for `json`. For example, in JupyterLab, there is a `json` viewer that nests properties nicely and such, but this by default will only work if the file extension is `json` (e.g. I renamed the one on the right to `.model`). 

image

Do you have a use-case in mind for why we would need it to be something more distinct?

lheagy:
I will merge this in as `discretize 0.4.9`. Feel free to move the conversation about file extensions for serialized meshes to new a issue @banesullivan - happy to continue the brainstorming 😄

discretize/utils/codeutils.py
```diff
@@ -1,16 +1,16 @@
...
+    scalarTypes += (long, )
+
 def isScalar(f):
-    if type(f) in scalarTypes:
+    if isinstance(f, scalarTypes):
         return True
-    elif isinstance(f, np.ndarray) and f.size == 1 and type(f[0]) in scalarTypes:
+    elif isinstance(f, np.ndarray) and f.size == 1 and isinstance(f[0], scalarTypes):
         return True
     return False
```
prisae, line 16: Sorry for commenting on an old PR @lheagy. I just stumbled upon that and I was wondering if there is any benefit of `discretize.utils.codeutils.isScalar(f)` (lines 6-15) over simply `np.size(f) == 1`? Just thought because you extended it for complex number, so it is fragile on the defined `scalarTypes` and will fail for all others (e.g., `int8`, `float32`, `complex64`, etc).

lheagy: No worries @prisae, thanks for raising this. In the `scalarTypes` we do have `np.number`  which `np.int8` and others inherit.

image

I think that just checking size isn't quite enough. For example, a 2D 1-element array currently returns `False` from isScalar as the first entry is an array. I _think_ that that logic is what we want. 

image

For example in the `TensorType` code, we should never be working with 2D arrays. 

https://github.com/simpeg/discretize/blob/master/discretize/utils/matutils.py#L299-L309

However, this might take a bit more thought. Maybe we do want anything that has one element to evaluate `True`?

prisae: Yes, it might be good to think about it and agree on that, and then write a little docstring with it. Because the behaviour of `isScalar` is not quite obvious from just looking at it.

Here some values

|                  | np.isscalar(x) | np.ndim(x) | np.size(x)==1 | isScalar(x) |
| ---------------- |:--------------:|:----------:|:-------------:|:-----------:|
| 0                | True           |     0      | True          | True        |
| np.array(0)      | False          |     0      | True          | IndexError  |
| np.atleast_1d(0) | False          |     1      | True          | True        |
| np.atleast_2d(0) | False          |     2      | True          | False       |
| np.atleast_3d(0) | False          |     3      | True          | False       |

`isScalar` currently fails for a 0D numpy array. However, instead of fixing `isScalar` for that case we might consider its purpose first.

What is the use-case that a one-element 1D array passes as scalar, but a one-element 2D or 3D array not?

prisae: Here the failure for the 0D numpy array:
```
In [58]: isScalar(np.array(0))                                                  
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-58-525f9e80cd80> in <module>
----> 1 isScalar(np.array(0))

~/anaconda3/lib/python3.7/site-packages/discretize/utils/codeutils.py in isScalar(f)
     11     if isinstance(f, scalarTypes):
     12         return True
---> 13     elif isinstance(f, np.ndarray) and f.size == 1 and isinstance(f[0], scalarTypes):
     14         return True
     15     return False

IndexError: too many indices for array
```

PR #185, opened 2019-08-14
Resolving #184

### LINKED ISSUE #184: Metaclass conflict when importing discretize
**What was done**
1. Using a new python 3.7.3 virtualenv, run `pip install discretize`
2. Start the python interpreter
3. run `import discretize`

**What was seen**
```
>>> import discretize
  File "<stdin>", line 1, in <module>
  File "/Users/ianmacaulay/tmp/venv3/lib/python3.7/site-packages/discretize/__init__.py", line 2, in <module>
    from discretize.TensorMesh import TensorMesh
  File "/Users/ianmacaulay/tmp/venv3/lib/python3.7/site-packages/discretize/TensorMesh.py", line 15, in <module>
    InnerProducts, TensorMeshIO
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

**What was expected**
Discretize should import properly.

OS: MacOS 10.14.6
python: 3.7.3
Version: 0.4.9

The only version that I was able to get to work was 0.4.4 on a clean install, although I was required to install numpy myself before I could install discretize

prisae:
The problem of #184 is that the plotting classes are inherited by other classes, so they should always be available. The fix is therefore to move the `@requires` from the class level down to the function level. This should answer @rowanc1 question :wink: (https://github.com/simpeg/discretize/pull/156#discussion_r281029700).

This PR will close #184.

### DIFF
--- a/discretize/TreeMesh.py
+++ b/discretize/TreeMesh.py
@@ -97,6 +97,15 @@ from scipy.spatial import Delaunay
 import scipy.sparse as sp
 from six import integer_types
 
+from discretize.utils.codeutils import requires
+# matplotlib is a soft dependencies for discretize
+try:
+    import matplotlib.pyplot as plt
+    import matplotlib
+except ImportError:
+    matplotlib = False
+
+
 class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
     """
     TreeMesh is a class for adaptive QuadTree (2D) and OcTree (3D) meshes.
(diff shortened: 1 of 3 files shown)


prisae:
Currently the soft-import is implemented in:

- Tests.py
- View.py

by doing a try-except at the top of the file
```
from discretize.utils.codeutils import requires
# matplotlib is a soft dependencies for discretize
try:
    import matplotlib
except ImportError:
    matplotlib = False
```

and add a `requires` before each function which needs matplotlib
```
@requires({'matplotlib': matplotlib})
```

Currently, this is not implemented in:
- TreeMesh.py
- tree_ext.pyx

We should probably take this PR to implement it for those too. Having all imports at the top of the file is a good idea anyway (and follows [PEP8](https://www.python.org/dev/peps/pep-0008/#imports): "*Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.*").

prisae:
Does it look OK for you regarding `TreeMesh.py` and `tree_ext.pyx` @jcapriot ?

prisae:
Great, thanks @jcapriot.

@lheagy (or someone with merging authorities), this is a bugfix for #184, and it seems to be good to go.

PR #189, opened 2019-10-07
Bugfix slicer

prisae:
Bugfix. Slicer failed if `xslice` was provided without a `yslice`, because of wrong comparison.

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -1346,6 +1346,10 @@ class Slicer(object):
         # Store min and max of all data
         if clim is None:
             clim = [np.nanmin(self.v), np.nanmax(self.v)]
+        # In the case of a homogeneous fullspace provide a small range to avoid
+        # problems with colorbar and the three subplots.
+        if clim[0] == clim[1]:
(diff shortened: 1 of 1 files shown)


prisae:
Thought I'd ping this one, as it is really a tiny change which is required and nothing happened for three weeks...

prisae:
I added another bugfix to the slicer. I'd be really happy if we could move these in at some point.

prisae:
Here graphically what this last bugfix fixes:

Before, if you provided a single value, it looked like this:
Selection_002
So the colour in the xy- and xz-plots is wrong. As soon as you scroll, the colour in the yz-plot will change to the wrong colour too.

With the bug-fix, they all get the right colour, and will remain correct if scrolling:
Selection_001

The bug only affects homogeneous fullspaces, which caused an issue with the range used by `colorbar()`.

prisae:
Thanks for reviewing/merging. I am the one to blame introducing them in the first place...

jcapriot:
Also means someone approved them in the first place... no worries

PR #191, opened 2020-01-09
Numpy build fix

### LINKED ISSUE #51: Crash on install of SimPEG seems to be caused by line in discretize/setup.py

### LINKED ISSUE #147: MyBinder pip install fails due to setup.py NumPy dependency
Does anyone know of a way to get around the NumPy dependency for installing `discretize` in a fresh environment?

It would be really awesome to have the `discretize` example notebooks on [Binder](http://MyBinder.org)! Currently, Binder servers with `discretize` fail to build if `discretize` is in the `requirements.txt` because of the numpy dependency.

### LINKED ISSUE #190: Colorbar in "Basic Forward 2D DC Resistivity"
https://discretize.readthedocs.io/en/docs_examples_clean/examples/plot_dc_resistivity.html

I didn't look into it in more detail, I simply stumbled upon this figure and realized there must be something wrong with the left colorbar, as it is jet, whereas the figure itself is viridis.

Selection_001

jcapriot:
This branch contains a few small bug fixes.
Mainly, it enables discretize to be installed on a system without NumPy present beforehand (for example when pip installing discretize). Resolve #51, Resolve #147
It also contains a small update to the curvilinear mesh view that updates that function to be in line with the other ``plotImage`` functions. Resolve #190 

Also note that this branch removes testing and automatic wheel building for python 2.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,7 @@ env:
 
 matrix:
   include:
-    # test the base code on 2.7, 3.6, 3.7
-    - name: "Python 2.7 - base"
-      os: linux
-      env:
-        - PYTHON=2.7
-        - TEST=true
-        - TEST_DIR="tests/base"
+    # test the base code on 3.6, 3.7
     - name: "Python 3.6 - base"
       os: linux
       env:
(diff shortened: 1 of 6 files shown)


prisae:
Thanks @jcapriot!

PR #193, opened 2020-01-15
Remove pymatsolver

prisae:
Remove `pymatsolver` from `build_requires` in `setup.py`, as it is not required (keep requirements low).


jcapriot:
This makes sense to me, it's already in the documentation that pymatsolver is recommended, and it isn't used anywhere within the discretize package, expect in a docstring.

Issue #197, opened 2020-04-06
Amount of cells not changing when changing Octree levels in utils.refine_tree_xyz


sarahgarre:
I am trying to generate a few Octree meshes for the same 2D geometry with increasing number of small cells. I was hoping just playing with the octree levels (adapt nr of cells for each level) would allow me to do that, but changing numbers there keeps on giving me the same end results.

```
ELEC_LENGTH = 0.13 #specific for this survey
POINT_ELEC_LOC = 0.6 * ELEC_LENGTH 
topo = np.c_[electrode_loc["x"], ( electrode_loc["z"] + POINT_ELEC_LOC )]
bump_2 = np.c_[ topo[0:9,0] - 1.6 , topo[0:9,1] ] 
bump_1 = np.c_[ topo[0:9,0] - 0.8 , topo[0:9,1] ]
bump_11 = np.c_[ topo[-9:,0] + 0.8 , topo[0:9,1] ]
bump_22 = np.c_[ topo[-9:,0] + 1.6 , topo[0:9,1] ]
topo = np.append(np.r_[bump_2,bump_1],topo,axis=0) #add extra bumps at the right
topo = np.append(topo,np.r_[bump_11,bump_22],axis=0) #add extra bumps at the left

SMA_CELL = .02
DOMAIN_SIZE = 50

expon = math.ceil(np.log2(math.ceil(DOMAIN_SIZE/SMA_CELL)))
NUM_CELL = (np.power(2,expon))
h = np.ones(NUM_CELL)*SMA_CELL 

#Create list to loop over tree levels
TreeList = [[9,9,3],
            [9,6,3],
            [6,6,3],
            [3,6,3],
            [3,3,3],]
meshList={}

for i,levels in enumerate(TreeList):
    print(i, levels)

    meshList[i] = Mesh.TreeMesh([h,h])# the origin is at the bottom left corner of the mesh. (default 0,0)
    meshList[i].x0 = [-40,-70]

    #Generate the refined Octree mesh
    meshList[i] = utils.refine_tree_xyz(
        meshList[i],topo, 
        octree_levels=levels,
        method='surface', 
        finalize=False) 

    # Add boundary between soil horizons to remove wavy pattern from topography in depth
    #x_refi = np.linspace(np.min(electrode_loc["x"]),np.max(electrode_loc["x"]))
    #y_refi = np.ones(x_refi.shape)* -.1
    #layer_refi = np.c_[x_refi,y_refi]
[...]

PR #198, opened 2020-04-07
Refine tree xyz

domfournier:
Addresses issue #197 
image

Cleanup and speed up the `surface2ind_topo` function from `simpeg.utils.modelutils`

image

### DIFF
--- a/discretize/base/base_mesh.py
+++ b/discretize/base/base_mesh.py
@@ -155,7 +155,7 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
         int
             number of nodes in the mesh
 
-        examples
+        Examples
         --------
         .. plot::
             :include-source:
--- a/discretize/base/base_mesh.py
+++ b/discretize/base/base_mesh.py
@@ -132,8 +132,8 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
         int
             number of cells in the mesh
 
-        Example
-        -------
+        Examples
+        --------
         .. plot::
             :include-source:
 
@@ -155,8 +155,8 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
         int
             number of nodes in the mesh
 
-        Example
-        -------
+        examples
+        --------
         .. plot::
             :include-source:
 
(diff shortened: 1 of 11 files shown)


domfournier:
@sarahgarre If you can try this branch on your side that would be great.

discretize/base/base_mesh.py
```diff
@@ -155,8 +155,8 @@ def nN(self):
         int
             number of nodes in the mesh
 
-        Example
-        -------
+        examples
```
lheagy, line 158:
```suggestion
        Examples
```

discretize/utils/meshutils.py
```diff
@@ -770,3 +771,113 @@ def inBox(cell):
...
+    """
+    if not isinstance(mesh, (discretize.TensorMesh, discretize.TreeMesh)):
+        raise TypeError("Mesh must be either TensorMesh or TreeMesh")
+
+    if grid_reference not in ["N", "CC"]:
+        raise ValueError("Value of grid_reference must be 'N' (nodal) or 'CC' (cell center)")
+
+    dim = mesh.dim - 1
+
+    if mesh.dim == 3:
```
lheagy, line 820: here, this won't work as is for a fully 3D cyl - mesh (because we use the `y`-dimension to capture `theta`). So we would need to instead grab the associated `cartesianGrid` (https://github.com/simpeg/discretize/blob/master/discretize/CylMesh.py#L1607-L1625) before any interpolation. 

For now, I would recommend just doing a quick check and raise a `NotImplementedError`

```
if isinstance(mesh, discretize.CylMesh) and not mesh.isSymmetric:
    raise NotImplementedError
```

lheagy: (it will be okay for the cylindrically symmetric meshes because there is no variation in y - which is what the test confirms)

docs/release/0.4.12-notes.rst
```diff
@@ -0,0 +1,26 @@
+.. currentmodule:: discretize
+
```
lheagy, line 2: Thanks for adding these! 🎉

fourndo:
Brilliant, thanks for this @jcapriot

PR #200, opened 2020-04-24
Refine tree xyz box

thast:
A PR building upon #198. Minor improvements of the `inBox` function, originally coded by @jcapriot to allow refinements up to a determined octree level (instead of always the finest.).

Note: This seems to slightly change the necessary input of the function (see modifications to the tests). I do not know if this is desirable. For that reason, I chose to create a new PR instead of happening it to the current PR #198 and you can decide what's best.

Edit, Additional notes: the change of behavior is affecting the outcome of the tutorial too. I think it can easily be fixed but I'd like to have your opinions on the syntax before spending time on it.


jcapriot:
I think we should probably update all of the functions to have the same behavior, and agree that being able to control explicitly which level things get refined to is probably a good (and expected) behavior.

jcapriot:
Specifically, the `inBall` function should be easy to update. Although the `surface` method will take a little more care.

domfournier:
Updated behaviour:

Refine to smallest + padding cells
image

Second level only + padding
image

Third level only + vertical pad only
image

fourndo (approved):
@domfournier

micmitch:
That looks great Dom! Thanks for updating this.

jcapriot:
closing this in favor of #237

PR #204, opened 2020-05-27
Remove restriction of sphinx < 2

prisae:
- The upcoming sphinx_rtd_theme v0.5.0 has a fix for the issue why we
  pinned it originally.
- numpydoc v1.0.0 was released, what also helps.
- sphinx is alread at version > 3, which I think is not that affected
  either.

Maybe wait with merging until v0.5.0 of sphinx_rtd_theme is out:
ttps://pypi.org/project/sphinx-rtd-theme


prisae:
sphinx-rtd-theme 0.5.0 is now released, so I think this can go in.

PR #211, opened 2020-07-02
Add ability to handle "norm" keyword in pcolorOpts on the TreeMesh

jcapriot:
The `TreeMesh.plotImage` function will now respect the `norm` option for the `pcolorOpts` dictionary.

This allows us to make better LogScaled model plots.


jcapriot:
Sure! let's deprecate it!

jcapriot:
so.... I deprecated a bit more than just that... I think I hit every plot kwarg that was camelCase

PR #212, opened 2020-07-13
Volume average

jcapriot:
This PR implements the volume averaging operation for ``TensorMesh`` and ``TreeMesh``, as well as combinations of the two. I have used the Tensor to Tensor myself, however the Tree to X operations could use some more test driving.

Of course there's some simple tests in here to ensure it's doing what I think it should be doing.

I was also running into issues getting the numpydoc's plot directive to automatically work in the Examples section of docstrings, and the solution was to remove the napoleon docs, since we're already using the `numpydoc` extension, and I couldn't find any examples of Google style docstrings, I just removed that from `conf.py`. So if this is okay, I'd like to just remove the `napoleon` Sphinx extension.


jcapriot:
@prisae I know I never responded to it, but I did see your comment before, and implemented something that does the same thing, but bypasses the numpy boolean indexing. Let me know if that is sufficient. The idea was to speed it up when you're interpolating a large mesh extent onto a smaller mesh extent, without loosing too much time.

prisae:
> @prisae I know I never responded to it, but I did see your comment before, and implemented something that does the same thing, but bypasses the numpy boolean indexing. Let me know if that is sufficient. The idea was to speed it up when you're interpolating a large mesh extent onto a smaller mesh extent, without loosing too much time.

I'll try to have a look before the meeting today.

prisae:
Yes, I like it now, great work @jcapriot!

jcapriot:
Yes, there is certainly a cost associated with the general Tensor to OcTree... but you do bring up a good point. There's no reason the function could not recognize that a Tensor and OcTree share the same base mesh and switch to the faster version.

domfournier:
YEa not sure why the first in the loop is different. Maybe some imports on the simpeg side?

lheagy:
@jcapriot: would it be possible to do a beta release to make this available on conda-forge? I am having trouble installing discretize from source on a remote machine at the moment...

discretize/utils/interputils.py
```diff
@@ -82,3 +83,109 @@ def interpmat(locs, x, y=None, z=None):
     Q = sp.csr_matrix((vals, (I, J)),
                       shape=(npts, np.prod(shape)))
     return Q
+
+def volume_average(mesh_in, mesh_out, values=None, output=None):
+    """Volume averaging interpolation between meshes.
```
lheagy, line 88: Thanks for such a beautiful docstring :)

jcapriot:
@domfournier Can you check the timings again for me? I added shortcuts for when meshes share the same base mesh

PR #217, opened 2020-07-29
Fix Slicer matplotlib-warning.

### LINKED ISSUE #215: matplotlib deprecated passing `norm` and `vmin/vmax` simultaneously.
> `discretize/View.py:1654:` MatplotlibDeprecationWarning: Passing parameters norm and vmin/vmax simultaneously is deprecated since 3.3 and will become an error two minor releases later. Please pass vmin/vmax directly to the norm when creating it.

This is potentially something we should take care for in `View()`.

prisae:
Closes #215

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -1344,9 +1344,8 @@ class Slicer(object):
         the shown range.
 
     clim : None or list of [min, max]
-        For `pcolormesh` (`vmin`, `vmax`).
-
-        Note that this overrules `vmin`/`vmax` provided in `pcolor_opts`.
+        For `pcolormesh` (`vmin`, `vmax`). Note: if you use a `norm` (e.g.,
+        `LogNorm`) then `vmin`/`vmax` have to be provided in the norm.
 
     xlim, ylim, zlim : None or list of [min, max]
         Axis limits.
@@ -1366,12 +1365,6 @@ class Slicer(object):
 
     pcolor_opts : dictionary
         Passed to `pcolormesh`.
-
-        Note 1: `vmin`/`vmax` provided in `pcolor_opts` are overruled by the
-                `clim`.
-
-        Note 2: If a `norm` is provided, then `vmin` and `vmax` should be
-                defined in the norm.
 
     """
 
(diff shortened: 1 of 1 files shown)


prisae:
What I am not sure about is... I think I included `clim` as a parameter because it is also there in `plotSlice` and `plotImage`. However, this way it can be provided as `clim` but as well in `pcolor_opts`. Not sure if it would be better to deprecate `clim`, and push the use of `pcolor_opts`, to make the discretize-code slimmer and leave these things to matplotlib?

What is your @jcapriot, @lheagy ?

prisae:
That needs more testing, because vmin/vmax are log if the norm is log... so the current version will fail. I think we need to talk about this, and maybe get rid of clim to simplify things.

prisae:
Good to go and ready for review. Works also with norms.

I made another small change. We should not patch matplotlib, and propagate deprecation warnings. We should simply not _cause_ matplotlib errors.

So now, these two possibilities run without warnings:
```
mesh.plot_3d_slicer(res, pcolor_opts={'norm': LogNorm(vmin=0.3, vmax=200)})
```
and
```
mesh.plot_3d_slicer(np.log10(res), clim=[np.log10(0.3), np.log10(200)])
```
and these two cases will raise a matplotlib deprecation warning, because they should be adjusted (to the above versions):
```
mesh.plot_3d_slicer(res, clim=[0.3, 200], pcolor_opts={'norm': LogNorm()})
```
and
```
mesh.plot_3d_slicer(res, clim=[0.3, 10], pcolor_opts={'norm': LogNorm(vmin=0.3, vmax=200)})
```


Here a minimum working example to test:
```
import warnings
import discretize
import numpy as np
from matplotlib.colors import LogNorm
from matplotlib.cbook import MatplotlibDeprecationWarning

# ALWAYS SHOW THE MATPLOTLIB WARNINGS
warnings.simplefilter('always', MatplotlibDeprecationWarning)

mesh = discretize.TensorMesh([5, 1, 1])
res = 10**np.arange(5)

# 1: Good: Provide vmin/vmax in the norm
mesh.plot_3d_slicer(res, pcolor_opts={'norm': LogNorm(vmin=1, vmax=1000)})

# 2: Good: Provide clim, but NO norm
mesh.plot_3d_slicer(np.log10(res), clim=[0, 3])

# 3: Deprecated: Provide clim with a norm (new deprecated since matplotlib 3.3) 
mesh.plot_3d_slicer(res, clim=[1, 1000], pcolor_opts={'norm': LogNorm()})

# 4: Deprecated: Provide clim with a norm AND norm-v{min;max} (new deprecated since matplotlib 3.3) 
mesh.plot_3d_slicer(res, clim=[1, 1000], pcolor_opts={'norm': LogNorm(vmin=1, vmax=1000)})
```

Issue #222, opened 2020-09-03
volume averaging between tree + tensor with different number of cells


lheagy:
I am running into an 
```ValueError: operands could not be broadcast together with shapes (33,) (29,) ```

When I am trying to perform the volume averaging between a tensor mesh and a tree mesh. The extent of the TreeMesh is smaller than the tensor, so I am not entirely sure where things are going awry

![image](https://user-images.githubusercontent.com/6361812/92148582-03c33680-edda-11ea-9b35-f37f1e51aff3.png)

Here is a minimal example to reproduce the error: 

```python
import numpy as np
import discretize

pf = 1.3
npad = 5
nc = 20
cs = 5

nc_tree = 32

mesh_tensor = discretize.TensorMesh([
        [(cs, npad, -pf), (cs, nc-2), (cs, npad, pf)],
        [(cs, npad, -pf), (cs, nc-2), (cs, npad, pf)],
        [(cs, npad, -pf), (cs, nc-2), (cs, npad, pf)],
    ],
    x0 = "CCC"
)

mesh_tree = discretize.TreeMesh([
        [(cs, nc_tree)], [(cs, nc_tree)], [(cs, nc_tree)],
    ],
    x0 = "CCC"
)

mesh_tree = discretize.utils.refine_tree_xyz(
    mesh_tree, np.c_[0, 0, 0], octree_levels=[8, 4, 2], method="radial", finalize=True
)

sigma = np.random.rand(mesh_tensor.nC)

sigma_tree = discretize.utils.volume_average(mesh_tensor, mesh_tree, sigma)
```

Issue #221, opened 2020-09-03
Syntax warnings in meshutils 


lheagy:
On the latest version of master, I get a number of Syntax warnings on import of discretize (running Python 3.8) 

```
/.../discretize/discretize/utils/meshutils.py:92: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if len(shape) is 1:
/.../discretize/discretize/utils/meshutils.py:94: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif len(shape) is 2:
/.../discretize/discretize/utils/meshutils.py:96: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif len(shape) is 3:
/.../discretize/discretize/TreeMesh.py:631: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if v_type is not 'CC':
```

jcapriot:
yup these should all get replaced!

PR #223, opened 2020-09-03
Syntax and bug fixes

jcapriot:
Patch for #221 and #222.

For some reason I was expecting allclose to return False if the arrays were of different length.

In addition, there are a few syntax fixes for comparisons to True/False which is generally frowned upon in pep8, and syntax fixes for comparisons to literals.

### DIFF
--- a/discretize/CylMesh.py
+++ b/discretize/CylMesh.py
@@ -1523,7 +1523,7 @@ class CylMesh(
         values = list(hang.values())
         entries = np.ones(len(values))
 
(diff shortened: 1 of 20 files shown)


lheagy (approved):
Thanks so much for all of this cleanup @jcapriot! One quick question - do you think it is worth updating one of the tests for the volume averaging to have meshes with different number of cells?

jcapriot:
Done, will get this merged in once passed

PR #226, opened 2020-09-22
matplotlib warnings: norm + vmin, vmax

lheagy:
update the logic in plotting so that we don't pass both a norm and vmin, vmax to resolve the Matplotlib warning: 

```
MatplotlibDeprecationWarning: Passing parameters norm and vmin/vmax simultaneously is deprecated since 3.3 and will become an error two minor releases later. Please pass vmin/vmax directly to the norm when creating it.
```

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -478,7 +478,18 @@ class TensorView(object):
             if clim is None:
                 clim = [v.min(), v.max()]
             v = np.ma.masked_where(np.isnan(v), v)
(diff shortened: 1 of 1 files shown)


prisae:
Along the lines of what I mentioned in #227 - I think it is good to mention why something (issue/PR) was closed, e.g., fixed in #xyz, it was decided it will not be fixed, this is outdated due to xyz, etc.

Users often stumble to (open or closed) GitHub issues and PR's, so it is good to guide them to the solution.

This issue was fixed in #227.

Issue #196, opened 2020-03-21
Add `__len__`, `size`, and `shape`


prisae:
Following up on a discussion on Slack: we should add `__len__`, `size`, and `shape` attributes. For a TensorMesh, e.g., `shape` is the same as `vnC`, and `size` is the same as `nC`. However, while `shape` and `size` is familiar to anyone working with numpy, `vnC` and `nC` are not.

prisae:
This was implemented in #227

Issue #218, opened 2020-08-19
matplotlib v3.3.0 deprecations


prisae:
Matplotlib v3.3.0 introduced many new features and deprecations, which will fail two minor versions later, see
- https://matplotlib.org/3.3.0/users/whats_new.html
- https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html

Several of these affect discretize (mainly `View.py`), and we have to take care of them sooner or later (currently they just generate a lot of warnings for users on matplotlib v3.3.0, but no errors). Amongst others:

- `pcolormesh` and `pcolor` have a new parameter `shading` (https://matplotlib.org/3.3.0/users/whats_new.html#pcolor-and-pcolormesh-now-accept-shading-nearest-and-auto).
- Parameters `norm` and `vmin`/`vmax` should not be used simultaneously (https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html#parameters-norm-and-vmin-vmax-should-not-be-used-simultaneously)

This list is likely to grow, I seem to stumble upon new warnings and errors frequently since I updated. I think the following does not affect discretize though:

- For `symlog`-scales, the keywords `linthreshx` and `linthreshy` have both be renamed to `linthresh`.
- `set_xticks(x)` and `set_xticklabels(labels)`: if x and labels have different dimensions it was ignored silently until now, but now it raises a ValueError.

prisae:
Related to https://github.com/simpeg/discretize/issues/215 and https://github.com/simpeg/discretize/pull/217

prisae:
I think you fixed these all (in #227 mostly, maybe in others too), right @jcapriot? I am closing this, please re-open if I am wrong.

PR #227, opened 2020-09-24
Restructure

jcapriot:
There are a lot of changes in here for deprecations. I will generate more details on our reasoning for this soon.
The main purpose is to move everything to a pep8 style of naming convention for modules, classes, methods, functions, and arguments.
These deprecations will likely stick around for a while (Current estimate is discretize v 1.0.0). So there should be plenty of time for users to update.
One big change is that the items such as vnC, or vnN now return tuples (immutable objects) as it is unsafe to actually change these. This was not guarded against previously. The downstream implication is that code that relied on these items being `numpy.ndarrays` will break. (i.e. things like `mesh.vnC + 1` will no longer work, instead something like `[n+1 for n in mesh.vnC]`) would likely be necessary.

I have also separated out the plot commands even further into a Matplotlib mixin style class for each mesh.


jcapriot:
I have tried to touch the testing as little as possible to ensure that the deprecations are functioning properly.

discretize/base/base_tensor_mesh.py
```diff
@@ -100,70 +118,56 @@ def __init__(self, h=None, x0=None, **kwargs):
...
-    def hz(self):
-        "Width of cells in the z direction"
-        return None if self.dim < 3 else self.h[2]
-
-    @property
-    def vectorNx(self):
+    def grid_nodes_x(self):
         """Nodal grid vector (1D) in the x direction."""
-        return np.r_[0., self.hx.cumsum()] + self.x0[0]
+        return np.r_[self.x0[0], self.h[0]].cumsum()
```
prisae, line 126: General discussion point: is `x0` descriptive, or would something like `origin` be better?

discretize/tensor_mesh.py
```diff
@@ -131,25 +135,25 @@ def _repr_html_(self):
...
-                self._vol = utils.mkvc(
-                    np.outer(utils.mkvc(np.outer(vh[0], vh[1])), vh[2])
+                self._vol = mkvc(
+                    np.outer(mkvc(np.outer(vh[0], vh[1])), vh[2])
                 )
         return self._vol
 
     @property
-    def areaFx(self):
+    def face_x_areas(self):
```
prisae, line 156: Same for the `{face;edge}_{x;y;z}_*`, are the `_{x;y;z}` required at all?

discretize/mixins/mpl_mod.py
```diff
@@ -0,0 +1,2070 @@
...
+from matplotlib.widgets import Slider
+import matplotlib.colors as colors
+import matplotlib.cm as cmx
+from matplotlib.collections import PolyCollection, LineCollection
+from matplotlib import rc_params
+from mpl_toolkits.mplot3d.art3d import Line3DCollection
+import matplotlib.transforms as mtransforms
+import discretize
+
+class InterfaceMPL(object):
```
prisae, line 19: This is so awesome @jcapriot !

discretize/mixins/mpl_mod.py
```diff
@@ -0,0 +1,2070 @@
+from __future__ import print_function
```
prisae, line 1: This is quite a big PR. I think it would make to get rid of all `from __future__ import` and `from six import` at once, what do you think?

discretize/mixins/mpl_mod.py
```diff
@@ -0,0 +1,2070 @@
...
+    """This class is used for simple ``discretize`` mesh plotting using matplotlib.
+
+    This interface adds three methods to the meshes. ``plot_grid`` will plot the
+    grid points of each mesh in 2D and 3D. ``plot_image`` for 2D image ploting of
+    models. and ``plot_slice`` for plotting a 2D slice through a 3D mesh.
+    """
+
+    def plot_grid(
+        self, ax=None, nodes=False, faces=False, centers=False, edges=False,
+        lines=True, show_it=False, **kwargs
```
prisae, line 29: I would vote to deprecate `show_it` - at least I fail to see the use-case (except for in the old-style gallery examples, but they are all going to change eventually, right?)

prisae: (This applies to all functions, not just `plot_grid`.)

discretize/mixins/mpl_mod.py
```diff
@@ -0,0 +1,2070 @@
...
+            kwargs['linewidth'] = kwargs.get('linewidth', rcParams['lines.linewidth'])
+
+        return plotter(
+            ax=ax, nodes=nodes, faces=faces, centers=centers, edges=edges,
+            lines=lines, show_it=show_it, **kwargs
+        )
+
+    def plot_image(
+        self, v, v_type='CC', grid=False, view='real',
+        ax=None, clim=None, show_it=False,
```
prisae, line 169: I would vote to deprecate `clim` everywhere, for a couple of reasons.
- matplotlib uses `vmin`/`vmax`, so it can be confusing
- It should really be provided in `pcolor_opts{'vmin': -1e-10, 'vmax': 1e10}`.
- If a norm is used, then it has to be provided in the norm, e.g., `pcolor_opts{'norm': LogNorm(vmin=1e-10, vmax=1e-5)}`

There are a couple of PR's (#226, #228) and issues (#218) related to this, so this big overhaul might be the right moment to deprecate `clim` entirely, IMHO.

discretize/mixins/mpl_mod.py
```diff
@@ -0,0 +1,2070 @@
...
+        else:
+            aspect1 = aspect
+            aspect2 = aspect
+        if aspect2 in ['auto', 'equal']:
+            aspect3 = aspect2
+        else:
+            aspect3 = 1.0/aspect2
+
+        # Get color limits.
+
```
prisae, line 1798: Feel free to copy paste my changes in #228 just in here and close that PR.

prisae (changes requested):
This is awesome work @jcapriot , I am really looking forward for this being merged. The entire package is much better structured and much easier to read, follow, and understand.

discretize/base/base_mesh.py
```diff
@@ -18,13 +20,31 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
     """
 
     _REGISTRY = {}
+    _aliases = {
+        'nC': 'n_cells',
+        'nN': 'n_nodes',
+        'nEx': 'n_edges_x',
+        'nEy': 'n_edges_y',
+        'nEz': 'n_edges_z',
```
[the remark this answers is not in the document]
craigmillernz: is it less inituitive what [0],[1],[2], are though? Are there cases when it wouldn't be x,y,z? Do I need to know that [0]=x etc?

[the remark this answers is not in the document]
prisae: Good point Craig. The main reason and I guess why we talked about it was because the meshes have SO MANY attributes, so it is a way to reduce the attribute-jungle.

But yes, everything is up for discussion. However, as we decided to do it this way for `hx`, `hy`, and `hz` etc it would be consistent. Of course, that very decision can also be discussed ;-)

I think it is clear, it is simply first dimension, second dimension, third dimension. They might not be aligned with your x, y, z.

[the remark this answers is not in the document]
craigmillernz: A note about what the 1st, 2nd, 3rd dimensions are in various situations (like cartesian or cylindrical axis) would be useful.  It's always frustrating when you have to guess what dimension the first slice corresponds to.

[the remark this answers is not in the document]
jcapriot: I would not ant to make `n_edges` a tuple quantity, It should inherently be a single integer that describes the total number of `edges` on a mesh, looking forward for unstructured meshes essentially. I do no like the fact that `nEx`, `nEy`, and `nEz` all live on `BaseMesh` currently. 

The meaning of `n_edges_x`(`nEx`) is the number of edges with an `x` (or first dimension) tangent direction.

Regards to notes about what the dimensions correspond to, I feel like that should inherently be left to whoever is using discretize. Normally we think of our `TensorMesh`s as the dimensions are (x, y, z)=(+East, +North, Up). However it is in fact valid fo ANY right handed coordinate system, so if you wanted to treat the dimensions as (x, y, z) = (+North, +East, Down), the returned matrices are still completely valid (Or even if you wanted to oddly rotate the coordinate systems).

discretize/mixins/mpl_mod.py
jcapriot: I agree. I had intended on doing this, but it slipped my mind. Thanks for the reminder!

discretize/mixins/mpl_mod.py
jcapriot: yeah, with all of the interfaces now having a common call, it'll be a lot easier to make the change consistent.

jcapriot:
Here is a listing of all deprecations that are currently implemented, with some of our reasonings. The moving of files around should not particularly impact the end user, as we have always import things from either the discretize module: `import discretize.TensorMesh` or from the discretize utils module: `from discretize.utils import ndgrid`. All of these style import will still work as before.

As far as property renaming, they fall under two categories, more explicit names with short aliases, and straight up deprecations. Aliases to some commonly used properties are still kept around, and will be noted in the more explicit name, a list of these are:

on `BaseMesh`:
* `n_cells=nC`
* `n_nodes=nN`
* `n_edges=nE`
* `n_faces=nF`
* `n_edges_x=nEx`, `n_edges_y=nEy`, `n_edges_z=nEz`
   - These quantities are not shape type properties, but the number of edges whose tangent is in that direction
* `n_faces_x=nFx`, `n_faces_y=nFy`, `n_faces_z=nFz`
   - As above these are also not properties defining the shape of anything, but the number of faces whose normal is in that direction

on `BaseRectangularMesh`:
* `shape_cells=vnC`
* `shape_nodes=vnN`
* `shape_edges_x=vnEx`, `shape_edges_y=vnEy`, `shape_edges_z=vnEz`
  - Unlike above, these are actually the shape of these quantities, which live on the mesh, and respectively `n_edges_x = prod(shape_edges_x)`.
* `shape_faces_x=vnFx`, `shape_faces_y=vnFy`, `shape_faces_z=vnFz`
  - Same comments as the `shape_edge_` properties.

on `BaseTensorMesh`:
* `grid_cell_centers=gridCC`
* `grid_nodes=gridN`
* `grid_edges_x=gridEx`, `grid_edges_y=gridEy`, `grid_edges_z=gridEz`
* `grid_faces_x=gridFx`, `grid_faces_y=gridFy`, `grid_faces_z=gridFz`
  - There was some discussion on the name for these properties, whether they should be called `locations_` instead of `grid_`. My original thought was that `grid` implied some sort of regular grid, but then was kindly reminded that there are also unstructured grids, and that made me a little more hesitant to change the name, even though `locations` is certainly more indicative of what these quantities are.

on `DiffOperators`:
* `average_face_to_cell=aveF2CC`, `average_face_to_cell_vector=aveF2CCV`
* `average_face_x_to_cell=aveFx2CC`, `average_face_y_to_cell=aveFy2CC`, `average_face_z_to_cell=aveFz2CC`
* `average_cell_to_face=aveCC2F`, `average_cell_vector_to_face=aveCCV2F`
* `average_edge_to_cell=aveE2CC`, `average_edge_to_cell_vector=aveE2CCV`
* `average_edge_x_to_cell=aveEx2CC`, `average_edge_y_to_cell=aveEy2CC`, `average_edge_z_to_cell=aveEz2CC`
* `average_node_to_cell=aveN2CC`, `average_node_to_edge=aveN2E`, `average_node_to_face=aveN2F`

on `TreeMesh`:
* Similar to above, all `n_hanging_XXX=nhX`, and `n_total_XXX=ntX`, for each `node=N`, `face_x,y,z=Fx,y,z`, and `edge_x,y,z=Ex,y,z`
* Also, similar to above: `grid_hanging_XXX=gridhX`, for each hanging `node=N`, `face_x,y,z=Fx,y,z` and `edge_x,y,z=Ex,y,z`

These aliases are also consistent across the discretize Meshes.

jcapriot:
Continuing for the Deprecations, which will be removed in version 1.0.0 of discretize:
The reasoning behind most of these is to change to a pep8 friendly name, and also to make the naming slightly more consistent. They also all throw FutureWarnings to notify end-users.

on `BaseMesh`:
* `normals` -> `face_normals`
*  `tangents` -> `edge_tangents`
*   `projectEdgeVector` -> `project_edge_vector`
*   `projectFaceVector` -> `project_face_vector`

on  `BaseRectangularMesh`:
* `r` -> `reshape`
* `nCx` -> `shape_cells[0]`, `nCy` -> `shape_cells[1]`, `nCz` -> `shape_cells[2]`
* `nNx` -> `shape_nodes[0]`, `nNy` -> `shape_nodes[1], `nNz` -> `shape_nodes[2]`
   -  These quantities might have led to confusion so to reduce name space clutter and make things consistent, they will be deprecated. The thought here is by looking at `nEx` and `nNx` you would not know that they represent two different quantities. One is the the number of edges whose tangents are in the x direction, the second is the number of nodes along the x-axis of the mesh. 

on `BaseTensorMesh`:
* `vectorNx` -> `grid_nodes_x`, `vectorNy` -> `grid_nodes_y`, `vectorNz` -> `grid_nodes_z`
* `vectorCCx` -> `grid_cell_centers_x`, `vectorCCy` -> `grid_cell_centers_y`, `vectorCCz` -> `grid_cell_centers_z`
    -  I'm still not completely sold on the names for the above two sets. I feel like they should also be named `grid_` so they are easily discoverable, but then they also do not have aliases yet. I also feel like the reasoning above for avoiding `nCx` and `nNx` would also apply here.
*  `getInterpolationMat` -> `get_interpolation_matrix`
*    `isInside` -> `is_inside`
*    `getTensor` -> `get_tensor`

on `MeshIO`:
* `readUBC` -> `read_UBC`
* `readModelUBC` -> `read_model_UBC`
* `writeUBC` -> `write_UBC`
* `writeModelUBC` -> `write_model_UBC`

on `DiffOperators`:
* cellGrad = cell_gradient
* cellGradBC = cell_gradient_BC
* `cellGradx` -> `cell_gradient_x`, `cellGrady` -> `cell_gradient_y`, `cellGradz` -> `cell_gradient_z`
    - These are all specifically the x, y, z component of the gradient operation
*  `nodalGrad` -> `nodal_gradient`
*   `nodalLaplacian` -> `nodal_laplacian`
*    `faceDiv` -> `face_divergence`
*    `faceDivx` ->`face_x_divergence`, `faceDivy` -> `face_y_divergence`, `faceDivz` ->`face_z_divergence`
    - Note how the ordering here is purposely different than the `cell_gradient_x,y,z` because the operation is taking the divergence of `face_x`, etc. There is no `x` component of a divergence operator because it is a scalar, there is an `x` component of the faces though.
*   `edgeCurl` -> `edge_curl`
*    `setCellGradBC` -> `set_cell_gradient_BC`
*    `getBCProjWF` -> `get_BC_projections`
*    `getBCProjWF_simple` -> `get_BC_projections_simple`

on `InnerProducts`:
* `getFaceInnerProduct` -> `get_face_inner_product`
* `getEdgeInnerProduct` -> `get_edge_inner_product`
* `getFaceInnerProductDeriv` -> `get_face_inner_product_deriv`
* `getEdgeInnerProductDeriv` -> `get_edge_inner_product_deriv`

jcapriot:
Some More deprecations for properties,

`CylMesh` -> `CylindricalMesh`

On `TensorMesh`, `CylindricalMesh`, `TreeMesh`, `CurvilinearMesh`:
* `vol` -> `cell_volumes`
* `area` -> `face_areas`
* `edge` -> `edge_lengths`

On `TensorMesh`, `CylindricalMesh`, `TreeMesh`:
*  `areaFx` -> `face_x_areas`, `areaFy` -> `face_y_areas`, `areaFz` ->`face_z_areas`
*  `edgeEx` -> `edge_x_lengths`, `edgeEy` -> `edge_y_lengths`, `edgeEz` ->`edge_z_lengths`

On `TensorMesh`, `TreeMesh`
* `faceBoundaryInd` -> `face_boundary_indices`
* `cellBoundaryInd` -> `cell_boundary_indices`

On `CylindricalMesh`:
* `isSymmetric` -> `is_symmetric`
* `cartesianOrigin` -> `cartesian_origin`
* `getInterpolationMatCartMesh` -> `get_interpolation_matrix_cartesian_mesh`
* `cartesianGrid` -> `cartesian_grid`

On `TreeMesh`:
* `maxLevel` -> `max_used_level`
* `permuteCC` -> `permute_cells`
*  `permuteF` -> `permute_faces`
* `permuteE` -> `permute_edges`

And for plotting with matplotlib:
* `plotGrid` -> `plot_grid`
* `plotImage` -> `plot_image`
* `plotSlice` -> `plot_slice`

discretize/base/base_mesh.py
```diff
@@ -124,274 +132,343 @@ def dim(self):
...
+        >>> mesh = discretize.TensorMesh([np.ones(n) for n in [2,3]])
+        >>> mesh.plot_grid(centers=True, show_it=True)
+        >>> print(mesh.n_cells)
         """
-        return int(self._n.prod())
+        return int(np.prod(self._n))
+
+    def __len__(self):
+        """The number of cells on the mesh."""
+        return self.n_cells
```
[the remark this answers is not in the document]
jcapriot: honestly, it comes up so much, and almost every time we define quantities of interest to live on cell centers, that it might as well just do this...

jcapriot:
I'm certainly open to moving more of the aliases to deprecations, specifically the old `gridCC` style things, and open to switching names up.

jcapriot:
Almost forgot the `discretize.utils` deprecations:

* `isScalar` -> `is_scalar`
* `asArray_N_x_Dim` -> `as_array_n_by_dim`
* `sdInv` -> `sdinv`
* `getSubArray` -> `get_subarray`
* `inv3X3BlockDiagonal` -> `inverse_3x3_block_diagonal`
* `inv2X2BlockDiagonal` -> `inverse_2x2_block_diagonal`
* `makePropertyTensor` -> `make_property_tensor`
* `invPropertyTensor` -> `inverse_property_tensor`
* `exampleLrmGrid` -> `example_curvilinear_grid`
* `meshTensor` -> `mesh_tensor`
   - I still would like a better name for this function. It takes the fancy input for the `h` of a tensor mesh, and expands it to the full cell width array`
* `closestPoints` -> `closest_points`
* `ExtractCoreMesh` -> `extract_core_mesh`
* `volTetra` -> `volume_tetrahedron`
* `indexCube` -> `index_cube`
* `faceInfo` -> `face_info`
* `interpmat` -> `interpolation_matrix`
* `rotationMatrixFromNormals` -> `rotation_matrix_from_normals`
* `rotatePointsFromNormals` -> `rotate_points_from_normals`

discretize/base/base_mesh.py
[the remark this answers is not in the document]
prisae: Fair enough. I remember we discussed it in the meeting when you first brought up the google docs. I thought it is worth thinking about it once again,  But sounds reasonable.

prisae (approved):

prisae:
Just an idea (and it works against the aim of reducing attributes...):

What if we introduce an `mtype` attribute, that works the same as the `dtype` attribute for NumPy arrays (hence mesh-type in analogy to data-type). So you could call
```
>>> mesh.mtype
mtype('TensorMesh')
```

Of course, there are other ways to achieve this now: `type(mesh)` yields `discretize.TensorMesh.TensorMesh`, and `mesh.__class__.__name__` yields `'TensorMesh'`, but at least the latter is not recommended, using private attributes.

It could be worth to check with @banesullivan if vtk or PyVista internally or other meshes have something like this.

banesullivan:
@prisae, I really like that idea of an `mtype`! VTK has an equivalent `GetClassName()` which we use internally in PyVista but perhaps we could adopt this `mtype` idea too.

banesullivan:
Also. @jcapriot: wow. this work is sersiously awesome! I won't have time/energy to provide any feedback but I am loving what I see here!

discretize/__init__.py
```diff
@@ -1,20 +1,19 @@
 # from discretize.BaseMesh import BaseMesh
-from discretize.TensorMesh import TensorMesh
-from discretize.CylMesh import CylMesh
-from discretize.CurvilinearMesh import CurvilinearMesh
-from discretize import Tests
-from discretize.MeshIO import load_mesh
+from .tensor_mesh import TensorMesh
+from .cylindrical_mesh import CylMesh, CylindricalMesh
+from .curvilinear_mesh import CurvilinearMesh
+from .base.mesh_io import load_mesh
```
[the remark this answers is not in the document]
jcapriot: I'm certainly happy to switch them all over, mostly I was just making things consistent as it was a mix of the two systems previously.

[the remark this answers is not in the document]
prisae: Personally I have no strong opinion. I just thought I'd mention it as the PR is about PEP8 conformity. But yes, I agree that more important is consistency than which one of them.

rowanc1 (commented):
Seriously impressed by all of the work that has gone into this PR, the updated names/structures and backwards compatibility is great.

Nice work @jcapriot.

discretize/__init__.py
```diff
@@ -25,9 +25,13 @@
...
+Tests = tests
 
-__version__   = '0.5.1'
-__author__    = 'SimPEG Team'
-__license__   = 'MIT'
-__copyright__ = '2013 - 2019, SimPEG Developers, http://simpeg.xyz'
+__version__ = "0.5.1"
+__author__ = "SimPEG Team"
+__license__ = "MIT"
+__copyright__ = "2013 - 2019, SimPEG Developers, http://simpeg.xyz"
```
prisae, line 37: Adjust (automate) year.

prisae: (I guess generally everything in #111, which includes the removing of `__future__` etc and the automation of the year)

discretize/base/base_mesh.py
```diff
@@ -18,19 +20,35 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
     """
 
     _REGISTRY = {}
+    _aliases = {
```
lheagy, line 23: nice, this is great!

lheagy: to @prisae's point on slack, do we want to add
```
"origin": "x0" 
``` 
to this list as well?

discretize/base/base_mesh.py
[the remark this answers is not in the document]
prisae: Closes #196

jcapriot:
my fix for the matplotlib 3.3 deprecation was to just to remove all internal passes of `vmin`, `vmax` to `pcolormesh` and conditionally add those two to the `pcolor_opts` dictionary if `clim is not None`. @prisae, @lheagy can you double check the deprecation warning is no longer thrown? (unless the user of course supplies both `clim` and `pcolor_opts['norm']`

prisae:
The way you implemented `clim` and took over #228 for the slicer does unfortunately not work. If you put `vmin` and `vmax` into `pcolor_opts`, but you use a `norm` (e.g., `LogNorm`), matplotlib will complain and in two minor releases fail. If a `norm` is used, `vmin` and `vmax` have to be provided in the norm, e.g., `pcolor_opts{'norm': LogNorm(vmin=vmin, vmax=vmax}`. Otherwise they have to be provided `pcolor_opts`: `pcolor_opts{'vmin': vmin, 'vmax': vmax}`. So just putting them into `pcolor_opts` all the time will lead to troubles.

Other then that, LGTM.

prisae:
For the `plot_slicer` I would prefer the way I provided in #228 , unless there are strong arguments against it.

jcapriot:
I would disagree. If we take steps to hide the issue, then we are committing the same error that `matplotlib` is intending to resolve. I specifically want the deprecation warning to be thrown if the user specifies a `norm` and a `clim`/`vmin,vmax`. I do not want it to throw an error if the user only supplies a `norm`, which is what I believe would happen previously, because we were internally using the `vmin`, `vmax` keywords to `pcolormesh`.

prisae:
It is not consistent in my opinion. So you can define a `clim` if you do not provide a `norm`, but you can't if you provide a `norm`. I agree we should not hide an issue, but I don't think this is the way. We can discuss it. What about simply deprecating `clim`? That would resolve that and be the cleanest.

prisae:
OK, I guess we discussed that today (clim/vmin/vmax), so I am fine with it. LGTM!

discretize/__init__.py
```diff
@@ -25,9 +25,10 @@
...
+from discretize import tests
 
-__version__   = '0.5.1'
-__author__    = 'SimPEG Team'
-__license__   = 'MIT'
-__copyright__ = '2013 - 2019, SimPEG Developers, http://simpeg.xyz'
+__version__ = "0.5.1"
+__author__ = "SimPEG Team"
+__license__ = "MIT"
+__copyright__ = "2013 - 2020, SimPEG Developers, http://simpeg.xyz"
```
lheagy, line 34: Should we just remove 2020 here? For the licensing, really only the initial year is needed (that way we don't have to keep updating it...)

discretize/base/base_mesh.py
```diff
@@ -124,274 +132,343 @@ def dim(self):
...
         """
         if self.dim < 3:
             return None
-        return int((self._n + np.r_[1, 1, 0][:self.dim]).prod())
+        return int(np.prod(x + y for x, y in zip(self._n, (1, 1, 0))))
 
     @property
     def vnE(self):
-        """Total number of edges in each direction
+        """The number of edges in each direction
```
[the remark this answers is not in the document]
lheagy: good question. Maybe `shape_edges` as the main property? but keeping `vnE` as an alias?

discretize/base/base_mesh.py
```diff
@@ -124,274 +132,343 @@ def dim(self):
...
-        return int((self._n + np.r_[0, 0, 1][:self.dim]).prod())
+        return int(np.prod(x + y for x, y in zip(self._n, (0, 0, 1))))
 
     @property
     def vnF(self):
-        """Total number of faces in each direction
-
-        :rtype: numpy.ndarray
-        :return: [nFx, nFy, nFz], (dim, )
+        """The number of faces in each direction
```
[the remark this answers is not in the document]
lheagy: `shape_faces`? as per above?

discretize/base/base_mesh.py
```diff
@@ -124,288 +125,375 @@ def dim(self):
...
-                np.zeros(self.nEz), np.zeros(self.nEz), np.ones(self.nEz)
+                np.zeros(self.n_edges_z),
+                np.zeros(self.n_edges_z),
+                np.ones(self.n_edges_z),
             ]
             return np.r_[tX, tY, tZ]
 
-    def projectFaceVector(self, fV):
-        """Given a vector, fV, in cartesian coordinates, this will project
+    def project_face_vector(self, fV):
```
lheagy, line 430: this is minor, but would it be worth `fV` --> `face_vector` to be a bit more explicit in naming

discretize/base/base_mesh.py
```diff
@@ -124,288 +125,375 @@ def dim(self):
...
+            len(fV.shape) == 2
+            and fV.shape[0] == self.n_faces
+            and fV.shape[1] == self.dim
         ):
-            raise Exception('fV must be an ndarray of shape (nF x dim)')
-        return np.sum(fV*self.normals, 1)
+            raise Exception("fV must be an ndarray of shape (n_faces x dim)")
+        return np.sum(fV * self.face_normals, 1)
+
+    def project_edge_vector(self, eV):
```
lheagy, line 457: similar to my comment above, perhaps `edge_vector` to simplify naming

discretize/base/base_mesh.py
```diff
@@ -416,38 +504,42 @@ def copy(self):
...
 
     @properties.validator
     def _validate_orientation(self):
         """Check if axes are orthogonal"""
-        if not (np.abs(self.axis_u.dot(self.axis_v) < 1e-6) and
-                np.abs(self.axis_v.dot(self.axis_w) < 1e-6) and
-                np.abs(self.axis_w.dot(self.axis_u) < 1e-6)):
-            raise ValueError('axis_u, axis_v, and axis_w must be orthogonal')
+        if not (
+            np.abs(self.axis_u.dot(self.axis_v) < 1e-6)
```
lheagy, line 526: here, it would be nice to set `tol = 1e-6` (I think it can stay inside the function, but just using a named variable for clarity of code)

discretize/base/base_mesh.py
```diff
@@ -466,324 +558,273 @@ def rotation_matrix(self):
...
+        **BaseMesh._aliases,
+        **{
+            "vnC": "shape_cells",
+            "vnN": "shape_nodes",
+            "vnEx": "shape_edges_x",
+            "vnEy": "shape_edges_y",
+            "vnEz": "shape_edges_z",
+            "vnFx": "shape_faces_x",
+            "vnFy": "shape_faces_y",
+            "vnFz": "shape_faces_z",
```
lheagy, line 618: ah, I see now that @prisae and my comments above would be addressed here. Do you think we should interchange which is the property and which is the alias?

discretize/base/base_tensor_mesh.py
```diff
@@ -22,20 +36,37 @@ class BaseTensorMesh(BaseMesh):
...
+        **{
+            "gridCC": "grid_cell_centers",
+            "gridN": "grid_nodes",
+            "gridFx": "grid_faces_x",
+            "gridFy": "grid_faces_y",
+            "gridFz": "grid_faces_z",
+            "gridEx": "grid_edges_x",
+            "gridEy": "grid_edges_y",
+            "gridEz": "grid_edges_z",
+        },
```
lheagy, line 52: I know we chatted a bit about this at the meeting, but just wanted to flag this as a potential item to revisit: do we want to prepend everything with `grid` or just have `cell_centers`, `nodes`, etc as the properties. As @jcapriot, `grid` implies "a framework of spaced bars that are parallel to or cross each other"

image

discretize/base/base_tensor_mesh.py
lheagy: +1 for origin (though I think having `x0` as an alias would still be appropriate)

discretize/base/base_tensor_mesh.py
```diff
@@ -341,62 +364,69 @@ def _getInterpolationMat(self, loc, locType='CC', zerosOutside=False):
...
+    ):
+        """Produces linear interpolation matrix
 
         Parameters
         ----------
         loc : numpy.ndarray
             Location of points to interpolate to
 
-        locType : str
+        loc_type : str
```
lheagy, line 426: would it be worth updating this to `location_type` to be a bit more explicit (since we are renaming anyways)

discretize/base/base_tensor_mesh.py
```diff
@@ -417,12 +447,24 @@ def getInterpolationMat(self, loc, locType='CC', zerosOutside=False):
...
+                FutureWarning,
+            )
+            zeros_outside = kwargs["zerosOutside"]
+        return self._getInterpolationMat(loc, loc_type, zeros_outside)
 
-    def _fastInnerProduct(
-        self, projType, prop=None, invProp=False, invMat=False
-    ):
-        """ Fast version of getFaceInnerProduct.
+    def _fastInnerProduct(self, proj_type, prop=None, inv_prop=False, inv_mat=False):
```
lheagy, line 466: thinking through names here too. What do you think of 
- `projection_type`
- `inverse_property`
- `inverse_matrix`

discretize/base/mesh_io.py
```diff
@@ -2,59 +2,60 @@
...
 
 try:
-    from .mixins import InterfaceTensorread_vtk
+    from ..mixins import InterfaceTensorread_vtk
 except ImportError:
     InterfaceTensorread_vtk = object
 
 
-def load_mesh(filename):
+def load_mesh(file_name):
```
lheagy, line 16: do we want the underscore here? or just `filename` ?

discretize/curvilinear_mesh.py
```diff
@@ -0,0 +1,437 @@
...
+        if len(change["value"][0].shape) != len(change["value"]):
+            raise ValueError("Dimension mismatch")
+
+    @property
+    def grid_cell_centers(self):
+        """
+        Cell-centered grid
+        """
+        if getattr(self, "_gridCC", None) is None:
+            self._gridCC = np.concatenate(
```
lheagy, line 110: to simplify maintenance down the line, I think having the private variable named the same as the public, just with the underscore would be helpful `self._grid_cell_centers`. Do you think there are any things we should be concerned about for doing that at this stage?

discretize/operators/differential_operators.py
```diff
@@ -1112,83 +1207,141 @@ def _aveN2Ey(self):
...
-        if getattr(self, '_aveN2F', None) is None:
+        if getattr(self, "_aveN2F", None) is None:
             # The number of cell centers in each direction
             if self.dim == 1:
                 self._aveN2F = self._aveN2Fx
             elif self.dim == 2:
-                self._aveN2F = sp.vstack((
-                    self._aveN2Fx, self._aveN2Fy
-                ), format="csr")
+                self._aveN2F = sp.vstack((self._aveN2Fx, self._aveN2Fy), format="csr")
```
lheagy, line 1300: similar comment as earlier about possibly renaming the private variables to be the same as the public ones

lheagy (commented):
This is looking really solid @jcapriot! Thank you for all of the cleanup!! 

I left a few minor comments, and I think there are a couple items we still might want to chat through
- Do we want `grid_cell_centers` or just `cell_centers`? 
- `origin` as a property? 

and then the last cleanup item that I think might be worth doing at this stage is renaming private variables to match the new public names. I am happy to help with this if you agree it is worth doing at this stage

discretize/__init__.py
jcapriot: Sure!

discretize/base/base_mesh.py
jcapriot: might as well just change it now

discretize/base/base_mesh.py
jcapriot: +1 for code clarity

discretize/base/base_tensor_mesh.py
jcapriot: as this a hidden internal function, it doesn't make much of a difference. However, might as well change them internally to prop_type -> projection_type, prop->property, inv_prop -> invert_property, inv_mat -> invert_output

jcapriot: or perhaps, `model` would be a better term than `property` as that is already a default python class

prisae:
Just a note for the future. I think we should avoid such massive PRs, and try to stick to one issue per PR. Hence this one would have been better, IMHO,
- 1 PR for name changes,
- 1 PR for azure pipelines,
- 1 PR for dropping Python 2.7 stuff,
- 1 PR for matplotlib fixes,
- etc,
and each of this PR should be **squashed** at the end, giving each commit a sensible commit message. I am trying to figure out what is causing my failures now due to this PR, and this is sort of difficult in a 234 commit PR with many cryptic commit messages... ;-) 

Not that I would follow these guidelines thoroughly, I blame myself here too, it is more something we should probably strive for...

PR #228, opened 2020-09-24
Fix for Slicer - clim/vmin/vmax - again

prisae:
Unfortunately, with #217 I re-introduced the bug reported here https://github.com/simpeg/discretize/pull/189#issuecomment-570761040 and fixed with #189.

The confusion exists probably because there are current three ways, how to define the limits of the colourbar:
- `clim`
- `vmin`/`vmax` in `pcolor_opts`
- `vmin`/`vmax` in `pcolor_opts['norm']`

This is not unique to the Slicer, plotSlice etc have the same.

One idea could be to deprecate the `clim` keyword, in favour of the matplotlib-native way with vmin/max.

### DIFF
--- a/discretize/View.py
+++ b/discretize/View.py
@@ -1464,30 +1464,39 @@ class Slicer(object):
             aspect3 = 1.0/aspect2
 
         # Get color limits.
+        # Currently there are three different ways, which is probably not good.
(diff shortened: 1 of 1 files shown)


prisae:
This was fixed in #227 instead.

PR #232, opened 2020-12-03
Exorcise Properties

jcapriot:
This PR will exorcise the dependance on properties for discretize.
Essentially any item that was a property is now an attribute.
The only mesh that is not valid on instantiation is the TreeMesh, which will be valid once it has been finalized.
Previous required keyword arguments are now positional arguments. (previous behavior passing these as keyword arguments is unchanged because python will pass keyword arguments that match positional arguments appropriately.)

There are still a few things that need to be ironed out (such as the load_mesh function, writing documentation of these attributes).

This should enable us to create clean numpy style documentation for classes!


discretize/mixins/vtk_mod.py
```diff
@@ -138,8 +138,11 @@ class InterfaceVTK(object):
        mesh.axis_u = (1,-1,0)
        mesh.axis_v = (-1,-1,0)
        mesh.axis_w = (0,0,1)
-       # Check that the referenc fram is valid
-       mesh._validate_orientation()
+       mesh.orientation = np.array([
+           [1, -1, 0],
```
lheagy, line 142: should this be build from `axis_u`, `v`, `w`?

jcapriot: I think it's better to set it each at once, so that it is guaranteed to be a valid orientation matrix, instead of assigning to each three, and then asking the user to validate them afterwards. It's possible for the mesh to get in a very odd state if they are individually set without validating them along the way.

The only other solution would be to require the user to set them explicitly in order, i.e. `axis_u`, `axis_v`, then `axis_w`.

jcapriot: Although, I guess you're just specifically referring to this instance in this example...

discretize/base/base_mesh.py
```diff
@@ -33,103 +31,164 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
...
-        required=True,
-    )
-
-    origin = properties.Array(
-        "origin of the mesh (dim, )",
-        dtype=(float, int),
-        shape=("*",),
-        required=True,
-    )
+    _items = {'shape_cells', 'origin', 'orientation', 'reference_system'}
```
lheagy, line 36: This looks good! A couple questions / thoughts on naming: 
- do we want to use a name that somehow indicates these are the properties that are serialized? e.g. `_serialized_properties` or something that indicates these are the set of properties that are necessary for re-creating the class? I don't feel strongly about this, just trying to explore if there is something that might be a touch more descriptive. 

- `reference_system` --> `coordinate_system` ?

discretize/base/base_mesh.py
```diff
@@ -647,6 +653,89 @@ def _parse_location_type(self, location_type):
...
+        """
+        .. deprecated:: 0.7.0
+          `axis_u` will be removed in discretize 1.0.0, it is replaced by
+          `mesh.orientation` for better mesh orientation validation.
+
+        See Also
+        --------
+        orientation
+        """
+        warnings.warn(
```
lheagy, line 667: It might be worth checking in on pyvista development here? If keeping these properties around simplifies interoperability, then it might be worth keeping them around?

discretize/tree_mesh.py
```diff
@@ -135,31 +133,26 @@ class TreeMesh(_TreeMesh, BaseTensorMesh, InnerProducts, TreeMeshIO):
             "gridhEz": "hanging_edges_z",
         },
     }
+    _items = {'h', 'origin', 'cell_state'}
```
lheagy, line 136: are `h` and `origin` defined on the BaseTensorMesh? It likely doesn't make too much of a difference if we want to inherit those or not, just wanted to raise it to ask

lheagy (approved):
This looks great @jcapriot! Thanks for all of your work to pull out properties. I left a few minor comments / questions. There isn't anything that I think blocks merging this, just clarifying a few things for myself and raising a couple small things to think through. Feel free to ping on here if there are other items that would be helpful to chat through!

discretize/tree_mesh.py
jcapriot: They are, but BaseTensorMesh also accepts the coordinate system and other kwarg parameters that are currently not supported on TreeMesh, so it was best to just restate the items here.

lheagy: :+1: makes sense. Thanks Joe!

discretize/base/base_mesh.py
```diff
@@ -33,103 +31,164 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
...
+
+        if orientation is None:
+            orientation = Identity()
 
-        super(BaseMesh, self).__init__(**kwargs)
+        self.orientation = orientation
+        if reference_system is None:
+            reference_system = "cartesian"
+        self.reference_system = reference_system
+        super().__init__(**kwargs)
```
prisae, line 64: What is the purpose of this line, `super().__init__(**kwargs)`?.

`BaseMesh` does not inherit any class, so I think this line can be deleted.

It is this line that causes my test with emg3d to fail. If I remove it everything works fine. It can of course be that I use inheritance wrongly in emg3d... but it appears strange to me to have a `super()` call when we are at the very bottom.

prisae: Ah, now I see. It is in the code because before we inherited from `properties.HasProperties` and `InterfaceMixins` - but now we do not inherit anything any longer, so it should go (I think).

prisae:
I left a comment in the code about the `super()` in `BaseMesh`.

Copying over from Slack to keep track of it:

> The only thing we could discuss about: make `to_dict` more general. Now it works for json. But, eg., if we would like to store to h5 a list of different lists (e.g, `h=[hx, hy, hz]`, if `hx`, `hy`, and `hz` are of different length) does not work very well. Saving three arrays `hx`, `hy`, `hz` might be more fail-proof in these circumstances. However, this would need a dedicated `from_dict` function as well.

discretize/base/base_mesh.py
```diff
@@ -33,103 +31,164 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
...
 
     def __getattr__(self, name):
         if name == "_aliases":
-            raise AttributeError  # http://nedbatchelder.com/blog/201010/surprising_getattr_recursion.html
+            raise AttributeError
         name = self._aliases.get(name, name)
-        return object.__getattribute__(self, name)
+        return super().__getattribute__(name)
```
prisae, line 70: I think it also applies to this `super()`, as there is no inherited class.

prisae: Running the test suite replacing `super()` with `self.` worked fine.

discretize/base/base_mesh.py
jcapriot: I'll put it up on the slack, but the purpose here is if you have multiple inheritance. If a derived class has multiple parents, the next super() call will be to the second parent. I'd have to look at how your doing this in your classes, as it could either be a use-case I hadn't thought of regarding backwards-compatibility, or you're not popping off keyword arguments when initializing parents.

jcapriot: So looking at your `TensorMesh` class in emg3d (https://github.com/emsig/emg3d/blob/1a3add15692492af5d9f007b7aac5cd83586289a/emg3d/meshes.py#L188), this `super()` call on `BaseMesh` ends up attempting to call the `__init__` method of `_TensorMesh`, which is probably not what you wanted. It would seem like its purpose is to be a fallback if discretize is not available, and it doesn't need to actually be inherited. It might be best to actually just conditionally define your `TensorMesh`, i.e. instead of defining an empty class at https://github.com/emsig/emg3d/blob/1a3add15692492af5d9f007b7aac5cd83586289a/emg3d/meshes.py#L45, define your own `dTensorMesh` there.

prisae: Yes, I could change
```
class TensorMesh(dTensorMesh, _TensorMesh):
```
to
```
class TensorMesh(dTensorMesh if dTensorMesh else _TensorMesh):
```

But I think this is a work-around for something introduced here that I believe should not be. IMHO, there should be no `super()` here.

discretize/base/base_mesh.py
```diff
@@ -33,103 +31,164 @@ class BaseMesh(properties.HasProperties, InterfaceMixins):
...
+                    for i, thing in enumerate(attr):
+                        if isinstance(thing, np.ndarray):
+                            attr[i] = thing.tolist()
+                out[item] = attr
+        return out
+
+    def __eq__(self, other):
+        if type(self) != type(other):
+            return False
+        return self.to_dict() == other.to_dict()
```
prisae, line 174: I think it is great to implement an `__eq__`, but the dict-comparison is flaky with regards to, e.g., floating point issues.

I haven't tested it, but I suspect the following will return `False` instead of `True`:
```
d1 = discretize.TensorMesh([[0.3,]], (0,))
d2 = discretize.TensorMesh([[0.2+0.1,]], (0,))
d1 == d2
```

discretize/base/base_mesh.py
prisae: What is the reason for `_items` being a dict, given they are only keys?

prisae: (...wouldn't it be better an immutable tuple than a mutable dict?)

jcapriot: This is not a `dict`, it's actually a `set`! (A set can only have unique entries)

prisae: Oh goodness me, of course, it was clearly too late yesterday :facepalm:

discretize/base/base_mesh.py
```diff
@@ -169,7 +169,7 @@ def to_dict(self):
                 out[item] = attr
         return out
 
-    def __eq__(self, other):
+    def equals(self, other):
```
prisae, line 172: Out of interest - what is the reasoning (and implication) of the rename? Would that mean that `meshA == meshB` could return something different than `meshA.equals(meshB)`?

jcapriot: It had to deal with hashing. Essentially it’s this: if an object implements `__eq__` it must also implement `__hash__` to be used as a key (and this happens quite often in SimPEG). Because our meshes are Mutable objects (I.e. `origin` can be changed) the hash value could change over its lifetime and this lead to undefined behavior. Thus we deemed it best to not worry about implementing those two and allowing default python behavior of equality and hashing (which default to `x is y` and `id(x)` respectively)

PR #234, opened 2020-12-23
Boundary Conditions

jcapriot:
I have a quick succinct implementation of Robin Boundary conditions for the `cell_gradient` matrix (currently just for TensorMesh). The function is as general as I could attempt to make it as it supports values for alpha, beta, and gamma defined as both constant values, or as vectors defined on each boundary_face. Essentially it uses a ghost cell to approximate the value and derivative at the boundary. Making it act directly on the `cell_gradient` matrix seemed the most straightforward method to me.

By supporting variable alpha, beta, gamma it can flexibly boundary conditions on that are different on every boundary. Say for a DC problem, we wanted zero-Neumann everywhere, but a simple zero-Dirichlet condition on the furthest cell from the source, this could be simply implemented.

The function returns two items, a Sparse Matrix A, and a vector b, such that the gradient operation with the given Robin boundary conditions can be represented as:
`` Gu = (mesh.cell_gradient + A)*u + b``
Formulating it this way does not require any modification to the divergence matrix, thus for a solution to a Laplacian like problem:
`` mesh.face_divergence @ ((mesh.cell_gradient + A)*u + b)`` It should be clear that in the solution phase, ``mesh.face_divergence@b`` should be subtracted from both sides of the equation.
Fortunately the matrix:
``sp.diags(-mesh.cell_volumes)@mesh.face_divergence @(mesh.cell_gradient + A)`` will still be a good ol' symmetric matrix, and will indeed be positive definite if A had any Dirichlet conditions.

- [ ] Add Tests


jcapriot:
Renaming this to "boundary conditions" as it has grown into something slightly more generalized.

jcapriot:
I've added all the tests for these (100% diff coverage). There's simple test in the tests/boundaries/ folder that show how to use this to solve PDE's with boundary conditions (on Poisson and Maxwell like problems). This should make for nice boundary condition possibilities for the SimPEG package across all of the simulations.

Essentially this adds missing averaging operators to faces. It also adds matrices and properties of the mesh that allow it to reference it's own boundaries (i.e. `mesh.boundary_faces`, `mesh.project_face_to_boundary_face`, `mesh.boundary_face_outward_normals`). The ordering of these items matches the order they appear in the full list of these quantities, (i.e. `boundary_nodes` are the specific subset of `nodes` that are on the boundary).

Future work possibilities:
* add on helper functions similar to the `cell_grad_weak_form_robin` operator that return the necessary matrix for simple operations, if we find ourselves repeating code.
* Extend formulation to the `CylindricalMesh` @lheagy!

lheagy (approved):
This looks great @jcapriot!

I appreciate the walk through in the tests. It would be nice to augment this with some tutorials (cc @dccowan :), but that can be a different PR.

PR #235, opened 2021-01-04
add xyzslice keyword to mesh.plot_slice

ckohnke:
The current implementation of mesh.plot_slice takes in an "ind" keyword to control what index along the normal axis is sliced. However, in physics simulations and models, one rarely cares about the index of the slice (say, y=index 16) and rather cares about the actual coordinate plane of the slice (say, y=600m) in the model. It is simple to convert between the index and actual coordinates in the mesh, but this is currently left as a task for the user in mesh.plot_slice. Other plotting methods (most notably mesh.plot_3d_slicer using the "xslice", "yslice", and "zslice" keywords) forgo the "ind" keyword in favor of using a coordinate slice, which is converted to an index slice for plotting.

This pull request adds the functionality to plot a coordinate plane to mesh.plot_slice. This is done by a new optional keyword in mesh.plot_slice. The keyword is "xyzslice" (I didn't like 'slice' because it is a python keyword, and using "xslice", "yslice", and "zslice" was too verbose. Open to suggestions.) and takes in a float. The float is then converted to an index using the same logic as mesh.plot_3d_slicer. The index is then used to plot the slice through the mesh as previous. 

Throws a warning if both "ind" and "xyzslice" keywords are used. Coordinate out of mesh behavior is the same as mesh.plot_3d_slice and returns either the 0 index or the max index along the axis, as appropriate. 

Only tested on scalar data using TensorMesh, but should work for vector data, and using a TreeMesh. Example included based on examples/plot_image.py


jcapriot:
Thanks @ckohnke, this looks great. Thanks for also thinking ahead about issuing a warning when the user specifies both an `ind` and a `xyzslice`. I'm inclined to "`slice_loc`" as the keyword name however.

ckohnke:
> Thanks @ckohnke, this looks great. Thanks for also thinking ahead about issuing a warning when the user specifies both an `ind` and a `xyzslice`. I'm inclined to "`slice_loc`" as the keyword name however.

I can make that keyword change. I also want to make an example/test with a tree mesh. I noticed a comment recently that mesh.plot_3d_slicer is only compatible with tensor meshes right now and want to make sure the indexing still holds.

jcapriot:
> > Thanks @ckohnke, this looks great. Thanks for also thinking ahead about issuing a warning when the user specifies both an `ind` and a `xyzslice`. I'm inclined to "`slice_loc`" as the keyword name however.
> 
> I can make that keyword change. I also want to make an example/test with a tree mesh. I noticed a comment recently that mesh.plot_3d_slicer is only compatible with tensor meshes right now and want to make sure the indexing still holds.

Interesting comment, I don't know why plot_3d_slicer wouldn't work on a `TreeMesh` now, specially because it now contains a unified call.

ckohnke:
`slice_loc` Seems to work with a `TreeMesh` (images below). 
tensor
tree

Found the comment again in `mpl_mod.py` line 667:

    if mesh_type != "tensor":
        raise NotImplementedError(
            "plot_3d_slicer has only been implemented for a TensorMesh"
        )

This might be a leftover and `plot_3d_slicer` actually does work with `TreeMesh`, or it uses a function `TreeMesh` doesn't have, but `TensorMesh` does. Here's a quick traceback and it looks like `TreeMesh` doesn't have `reshape`, but `TensorMesh` does. Not sure I'm following the full logic behind how `plot_3d_slicer` is getting the 3 slices and then plotting them.

```
  File "plot_slice_loc.py", line 121, in <module>
    run()
  File "plot_slice_loc.py", line 118, in run
    O.plot_3d_slicer(ov)
  File "../discretize/mixins/mpl_mod.py", line 685, in plot_3d_slicer
    tracker = Slicer(
  File "../discretize/mixins/mpl_mod.py", line 2233, in __init__
    self.v = mesh.reshape(v.reshape((mesh.nC, -1), order="F"), "CC", "CC", "M")
  File "../discretize/base/base_mesh.py", line 75, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'TreeMesh' object has no attribute 'reshape'`
```

jcapriot (dismissed):
LGTM! Thanks again @ckohnke, this will be a nice quality of life improvement.

jcapriot:
@ckohnke Actually, the examples/plot_xyzslice.py should be removed, then it'll be good.

jcapriot:
@ckohnke I adapted the example you wrote into the docstring of the function, thanks for that!

PR #236, opened 2021-02-02
Correct the range_y in Tree plot_image

Rockpointgeo:
In the TreeMesh plot_image method defining the range_y does not result in the expected behaviour.

Instead of using range_y to define the limits of the plot it uses range_x.

I didn't mean to include the change to my .gitignore file in the pull request but don't know how to remove it now.


jcapriot:
Thanks for catching this, I've merged the master branch in, and removed the .gitignore changes for you. We'll get this in as soon as the testing goes through.

jcapriot (approved):

PR #237, opened 2021-02-10
Tree refine c

jcapriot:
Add's C implementation of a couple TreeMesh refine functionality

1) Refine all cells within a ball/s to a certain level/s
2) Refine all cells within a box/es to a certain level/s


leonfoks:
More of a comment, but you might find that cells may or not refine because of the numerical roundoff of the r = (x-x0)^2 + (y-y0)^2 + (z-z0)^2 term i.e. the split criteria (r_test >= r2) is very close to zero.

If you notice this, you will need to implement Shewchuk's geometric predicates  https://people.eecs.berkeley.edu/~jrs/papers/robustr.pdf.  It's a relatively small increase in time for an exact computation and only goes exact for those scenarios that need it.

It looks like someone has this online already in c! https://github.com/danshapero/predicates
But they only did the base predicates, not the in_circle/in_sphere test, but the paper shows exactly which ones to use for that and I did it here if thats easier to follow (but i doubt it ^_^) https://github.com/leonfoks/coretran/blob/master/src/spatial/sm_geometry.f90

jcapriot:
Cool, that worked for included Cython files in the coverage report!

PR #240, opened 2021-03-18
Tree quiver plot

jcapriot:
Add capability to plot vectors on the TreeMesh using Matplotlib's quiver on plot_image and plot_slice.
Inspired by Dom's plotting functions in the SimPEG examples.

the keyword argument to control options passed on to `quiver` is `quiver_opts`


ngodber:
....and then I see this.

PR #251, opened 2021-06-07
refactor closest_points_index to use cKDTree. This yields significant…

ngodber:
This is my working branch for improvements to discretize. First one is an easy one, but yields significant performance improvement.

### DIFF
--- a/discretize/base/base_mesh.py
+++ b/discretize/base/base_mesh.py
@@ -5,9 +5,9 @@ Base classes for all discretize meshes
 import numpy as np
 import os
 import json
-
+import scipy.spatial
 from discretize.utils import mkvc, Identity
-from discretize.utils.code_utils import deprecate_property, deprecate_method
+from discretize.utils.code_utils import deprecate_property, deprecate_method, as_array_n_by_dim
 import warnings
 
 
@@ -619,6 +619,31 @@ class BaseMesh:
         items.pop("__module__", None)
         items.pop("__class__", None)
         return cls(**items)
+
+    def get_nearest_indices(self, locations, grid_location='CC'):
+        """Return nearest indices to points.
+
+          Parameters
+          ----------
+          pts: numpy.ndarray
+              Points to move
+          grid_loc: {'CC', 'N', 'Fx', 'Fy', 'Fz', 'Ex', 'Ex', 'Ey', 'Ez'}
+              Which grid to move points to.
+
+          Returns
+          -------
+          numpy.ndarray
+              nodeInds
+          """
+
+        tree_name = f'_{grid_location}_tree'
+        pts = as_array_n_by_dim(locations, self.dim)
+        if getattr(self, tree_name, None) is None:
+            grid = getattr(self, "grid" + grid_location)
+            setattr(self, tree_name, scipy.spatial.cKDTree(grid))
+        tree = getattr(self, tree_name)
+        _, ind = tree.query(pts)
+        return ind
 
     @property
     def reference_is_rotated(self):
--- a/discretize/utils/curvilinear_utils.py
+++ b/discretize/utils/curvilinear_utils.py
@@ -220,6 +220,8 @@ def face_info(xyz, A, B, C, D, average=True, normalize_normals=True, **kwargs):
     return N, area
 
 
+
+
 volTetra = deprecate_function(volume_tetrahedron, "volTetra", removal_version="1.0.0")
 indexCube = deprecate_function(index_cube, "indexCube", removal_version="1.0.0")
 faceInfo = deprecate_function(face_info, "faceInfo", removal_version="1.0.0")
--- a/discretize/utils/mesh_utils.py
+++ b/discretize/utils/mesh_utils.py
@@ -228,17 +228,10 @@ def closest_points_index(mesh, pts, grid_loc="CC", **kwargs):
 
     pts = as_array_n_by_dim(pts, mesh.dim)
     grid = getattr(mesh, "grid" + grid_loc)
-    nodeInds = np.empty(pts.shape[0], dtype=int)
+    tree = cKDTree(grid)
+    _, ind = tree.query(pts)
 
-    for i, pt in enumerate(pts):
-        if mesh.dim == 1:
-            nodeInds[i] = ((pt - grid) ** 2).argmin()
-        else:
-            nodeInds[i] = (
-                ((np.tile(pt, (grid.shape[0], 1)) - grid) ** 2).sum(axis=1).argmin()
-            )
-
-    return nodeInds
+    return ind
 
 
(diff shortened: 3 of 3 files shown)


ngodber:
@jcapriot not sure exactly what is going on. Something to do with doc building? All the mesh tests pass and no modification was made that involved the named modules

`2021-09-14T10:26:16.4325660Z` [91mWARNING: [autosummary] failed to import 'discretize.base.mesh_io.TensorMeshIO': no module named discretize.base.mesh_io.TensorMeshIO[39;49;00m
2021-09-14T10:26:16.4337310Z [91mWARNING: [autosummary] failed to import 'discretize.base.mesh_io.TreeMeshIO': no module named discretize.base.mesh_io.TreeMeshIO[39;49;00m`

ngodber:
Tracked down source of test doc fails. Long story short it is a regression upstream, https://github.com/jupyter/nbconvert/issues/1430

Detailed fail is:

`sphinx-build -b linkcheck -d _build/doctrees   . _build/linkcheck
Running Sphinx v4.2.0
[autosummary] generating autosummary for: /home/master/discretize/docs/api/index.rst
[autosummary] writing to /home/master/discretize/docs/api/generated
WARNING: [autosummary] failed to import 'discretize.base.mesh_io.TensorMeshIO': no module named discretize.base.mesh_io.TensorMeshIO
WARNING: [autosummary] failed to import 'discretize.base.mesh_io.TreeMeshIO': no module named discretize.base.mesh_io.TreeMeshIO

Exception occurred:
  File "/usr/lib/python3.8/pathlib.py", line 1198, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/usr/share/jupyter/nbconvert/templates/conf.json'`

full traceback:

'Traceback (most recent call last):
  File "/home/master/.local/lib/python3.8/site-packages/sphinx/cmd/build.py", line 276, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/home/master/.local/lib/python3.8/site-packages/sphinx/application.py", line 237, in __init__
    self.setup_extension(extension)
  File "/home/master/.local/lib/python3.8/site-packages/sphinx/application.py", line 393, in setup_extension
    self.registry.load_extension(self, extname)
  File "/home/master/.local/lib/python3.8/site-packages/sphinx/registry.py", line 429, in load_extension
    mod = import_module(extname)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/master/.local/lib/python3.8/site-packages/nbsphinx.py", line 292, in <module>
    """.replace('__RST_DEFAULT_TEMPLATE__', nbconvert.RSTExporter().template_file)
  File "/home/master/.local/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 330, in __init__
    super().__init__(config=config, **kw)
  File "/home/master/.local/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 114, in __init__
    self._init_preprocessors()
  File "/home/master/.local/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 497, in _init_preprocessors
    conf = self._get_conf()
  File "/home/master/.local/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 517, in _get_conf
    if conf_path.exists():
  File "/usr/lib/python3.8/pathlib.py", line 1407, in exists
    self.stat()
  File "/usr/lib/python3.8/pathlib.py", line 1198, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/usr/share/jupyter/nbconvert/templates/conf.json'

jcapriot:
On `TreeMesh` the corresponding function is called `point2index`. However that only finds the containing cell.

The fastest implementation for tensor based meshes will be to use numpy's `searchsorted` function and then `sub2ind` to return the index. No need to build a full `KDTree` for these nicely structured meshes.

Also, in recent versions of `scipy`, `cKDTree` and `KDTree` are the same thing.

ngodber:
Yes agree - search sorted would be a bit faster but less general in that it would only work for tensor mesh, (not curvilinear or tree or in the future, triangular) and even for cases where we include the tree build I'm not sure it would be materially faster.  Storage of the tree substantially ameliorates the cost of building it and is a key feature. More importantly, as is it is a ~100x improvement in performance compared to current implementation if someone cares enough about the further improvements possible by handling tensor mesh specially they can add the logic.  

Point2Index assumes users want containing cell which is not necessarily the case. This function simply does what it says on the tin (say if you care more about discretization error).

ngodber:
@jcapriot point2index is Treemesh specific? Is there a general mesh function for the other types? My understand was this function (formerly in mesh.utils) was it. Not storing the tree really bogs down performance in any case where the function is called within a loop (as is the case in res). Of course you could refactor calls to remove loops but that breaks encapsulation and requires users to make sure to do things the 'Right Way'.  I'm surprised at the reluctance to treat it as a property, it is only created and stored on demand and memory wise is not burdensome? In cases where we are creating 100's or 1000's of mesh I could see it being material, in which case store=False is an option. 

@domfournier This is being used in the 2.5D res simulations where mesh.utils was being used to find the locations to place the source. That part of the code (building the RHS of the matrix) was very slow, often taking 80% of the run time. Switching to KDTree was faster but still slow as the tree was being rebuilt on every loop. Hence I was prompted to take the @property approach.  Performance improvements should be across any code where multiple source locations are found on the same mesh.

jcapriot:
I'm definitely not opposed to pulling it in, just wanting to make sure it's _right_ to pull it in, if that makes sense.

As a general question, how often have you used this to find the nearest index to something besides the cell center?

ngodber:
@jcapriot Off the top of my head I'm not using it in code outside of SimPEG itself. I am using it in https://github.com/simpeg/simpeg/pull/1007 . The implementation is identical between the mesh property method and utils so it is used in https://github.com/simpeg/simpeg/search?p=1&q=closestPoints. I'd infact forgotten that I had retained the utils implementation. I would additionally advocate a deprecation warning in utils referencing moving to the mesh property and change current code references to the new method. I'd also rename get_nearest_indices to closest_points_index.

ngodber:
@jcapriot merged. Liked the changes and better documentation. Any chance of getting into 0.7.1?

ngodber:
@jcapriot tests passed - good to merge.

PR #253, opened 2021-06-18
Numpy docstrings api

dccowan:
Pull Request for Discretize Docstrings
========================

- Global reorganization of autogenerated API
- Numpy style docstrings for each property, method, function and class
- Include examples within docstrings whenever possible/necessary

Some remaining items
-------------------------
- **How much information do we want to put in docstrings for differential operators?**
    - How much mathematical detail?
    - How much detail regarding how the operator is constructed?
    - What type of image or example to illustrate?


jcapriot:
Hey devon, this is looking good so far. I'm planning to go through this in much more detail tomorrow, but first off I'll make a few formatting requests.

1. If a function works with an arbitrary array class, use the term "array_like" (I believe I got this term set up to automatically link to the numpy.ndarray class)

2. When referring to numpy arrays (i.e. when returning items), use the term "numpy.ndarray" as this will properly cross link to that class

3. When describing shapes of arrays, lets be consistent and use the format in scipy's documentation, i.e. check out https://github.com/scipy/scipy/blob/v1.7.0/scipy/linalg/basic.py#L40-L261. I know I've probably put in a few things that do not follow this, so a pass through updating this would be great.

4. For all of the examples within the documentation, I'd like to use the format of 
```
>>>
>>>
>>>
```
(put commented lines outside the source code)
```
>>> 
>>>
```

This also means changing all of the `..plot_it:` sections to be like this as well. To trigger plotting, you just need to put in a line of either
```
>>> import matplotlib
```
or
```
>>> from matplotlib import pyplot as plt
```
at some point in the example

jcapriot:
Also another note, you seem to have removed some instances of the outputs after examples, (i.e. the `TreeMesh.refine()` function). Please leave this in, and add them to the other examples as well. Ideally the lines just after the examples will match the output of the code if a user ran that block.

dccowan:
@jcapriot are the 'stencils' for the operators meant to be hidden or visible? You have **_stencil_nodal_gradient** but you also have **stencil_cell_gradient** . I don't see any reason to make the stencils visible but maybe I am mistaken

image

jcapriot:
The cell gradient stencils are useful for designing regularizations, and also for potentially identifying faces associated with changes between cells.

dccowan:
> The cell gradient stencils are useful for designing regularizations, and also for potentially identifying faces associated with changes between cells.

Okay. So do we want them to be visible (no "_") then? If so, we should make all stencil functions visible.

dccowan:
I was documenting inner products and found that the return for 'edge_inner_product_deriv' isn't actually a sparse matrix. It is something a bit more abstract. I understand its purpose and I have some idea of how it could be implemented, but I was hoping to form it explicitly to better understand it before documenting.

dccowan:
@jcapriot I am ready for a review of the documentation for *Differential Operators* and *Inner Product* classes. I have done the documentation in differential_operators.py and inner_products.py. Once we are happy, can update the docstrings for these methods/properties when they are overwritten for another mesh class (e.g. tensor mesh).

PR #258, opened 2021-08-04
Numpy docstrings api review

jcapriot:
These are mostly a bunch of formatting edits

A few things I've noticed:

Let's stick to the SciPy convention of describing array/sparse matrix shapes, i.e. put them before the type:
`(N) numpy.ndarray` for single dimension arrays don't use stray comma.
`(N, M) numpy.ndarray` for higher dimensions DO separate with commas.
Also as part of this, do not put in the description that an item is an array with a certain shape, because that should be explicit from the type definition.

`numpy.ndarray` is the name used to specifically refer to the numpy array class everywhere. If a function accepts anything that looks like an array, use `array_like`, if a function accepts an iterable object, there is also a keyword for that as well (`iterable`). There is also a keyword `scalar` that works similar to `array_like`.

Do not use blank lines in example code blocks, i.e. `>>>` on its own is not good. Instead just break the example block where appropriate with two blank lines.

Use `...` for when a code block input spans multiple lines, (this should look like python's command prompt).

When intending for there to be a plot at a certain point in the examples, use `plt.show()` to place it after that code block.

Do not have comment lines within code blocks. These look cleaner in the documentation as rendered text and will be placed into comments anyways when a user exports the code for the example.

When trying to show text output of a code block, place the desired output directly on the next line, without a `>>>`.

Avoid stray blank lines at the end of the docstring, or at the end of a section.

There is no need to document the first argument of a class method (it is similar to `self`).

Be aware of the 80 character line limit, there were a few cases where the line lengths were crazy long (especially when there were lists).

There might be others that I did not elaborate on though.


jcapriot:
@dccowan Just want to make sure you see all the changes here.

jcapriot:
I have added class inheritance diagrams to this as well now, (this does require graphviz be installed on the documentation build host though). Fortunately conda and conda-forge both have recipes for them on every system.

dccowan:
I was able to build the docs with your current changes. I will go through again and finish up what you started.

PR #262, opened 2021-08-19
Fix wrong colour for fullspaces - again

### LINKED ISSUE #261: Regression 3d_slicer
I _think_ this is a regression. I remember this issue existed before, and I did fix it (pre 0.6.0). However, I cannot find it now.

Issue: if you plot a fullspace, the z-y-plot has a different colorscale.
```
import discretize
import numpy as np
mesh = discretize.TensorMesh(([1, 1], [1, 1], [1, 1]), (0, 0, 0))
mesh.plot_3d_slicer(np.ones(mesh.n_cells))
```
example

prisae:
Closes #261

### DIFF
--- a/discretize/mixins/mpl_mod.py
+++ b/discretize/mixins/mpl_mod.py
@@ -2379,9 +2379,10 @@ class Slicer(object):
             if clim[0] == clim[1]:
                 clim[0] *= 0.99
                 clim[1] *= 1.01
-        else:
-            self.pc_props["vmin"] = clim[0]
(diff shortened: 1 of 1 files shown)


prisae:
> Sorry for taking a while to get to this, Looking at it, I think this will cause a matplotlib `DeprecationWarning` if the user gives a norm and a clim, so I think it'll work best to instead remove instead put the "`else`" case after line. I've marked it in the review comment.

Thanks, good catch!

PR #272, opened 2022-02-20
Add dot product test.

prisae:
## Adds the dot product test to the test suite

I was thinking of adding the dot test to `emg3d`, but thought that maybe it lives better in `discretize`, alongside the `OrderTest` and `check_derivative`. As such it is available to any tool that uses `discretize`, including `SimPEG`.

In `SimPEG` there are dozens of places where the dot-test is defined. I think it would be better to only define it once, here in `discretize`, and use always the same definition (avoiding code repetitions and slightly different implementations). However, adding it in `discretize` means that we would probably have to wait at least for 2 minor releases before we can use it in `SimPEG`.

I draw a lot of inspiration from the implementation in PyLops, and we are reworking the `dottest` over there currently (https://github.com/PyLops/pylops/pull/335, https://github.com/PyLops/pylops/pull/336).

Suggestions for further improvements or adaption for `SimPEG`/`discretize`-language/habits/conventions are welcomed!

### Open questions to discuss

- [ ] Defaults for `rtol`, `atol`?
- [ ] Should we have `shape` and `order` variables? Alternatively, we could only support size, and if a particular shape/order is required one has to wrap it like
   ```
   x, y = 10, 20
   dottest(
       fwd=lamda u: my_fwd_op(u.reshape((x, y), order='F')).ravel('F'),
       ...
       fwd_size=x*y,
       ...
   )
   ```
   So it would be possible even when just having `fwd_size` instead of `fwd_shape`, and no `fwd_order` (all the same applies of course to `ajd_...`). But having it inside the functions makes the `dottest` a bit easier to use for various scenarios.
- [ ] Default order is set to `"C"` - would it, with `discretize`, make more sense to set the default order to `"F"`? Or are most operators on 1D vectors anyway?

### DIFF
--- a/discretize/tests.py
+++ b/discretize/tests.py
@@ -637,9 +637,21 @@ def get_quadratic(A, b, c=0):
     return Quadratic
 
 
-def dottest(forward, adjoint, fwd_size, adj_size, fwd_complex=False,
-            adj_complex=False, rtol=1e-6, atol=0.0, raise_error=False,
-            verb=False):
+# TESTS
+def dottest(
+    forward,
+    adjoint,
+    fwd_shape,
+    adj_shape,
+    fwd_complex=False,
+    adj_complex=False,
+    fwd_order="C",
+    adj_order="C",
+    rtol=1e-6,
+    atol=0.0,
+    raise_error=False,
+    verb=False,
+):
     r"""Dot product test for the forward operator and its adjoint operator.
 
     Dot product test to verify the correctness of the adjoint operator
@@ -656,11 +668,15 @@ def dottest(forward, adjoint, fwd_size, adj_size, fwd_complex=False,
     forward, adjoint : functions
         Forward operator and its adjoint operator.
 
-    fwd_size, adj_size : int
-        Sizes of the vectors {u; v} passed to ``forward`` and ``adjoint``.
+    fwd_shape, adj_shape : int, tuple of int
+        Shapes of the vectors {u; v} passed to ``forward`` and ``adjoint``.
 
     fwd_complex, adj_complex : bool, defaults: False, False
         If True, complex vectors are passed to ``forward`` and ``adjoint``.
+
+    fwd_order, adj_order : string, defaults: 'C', 'C'
+        Used the ravel the output of the ``forward``/``adjoint`` if their shape
+        is higher than 1D.
 
     rtol, atol : float, defaults: 1e-6, 0.0
         Relative and absolute tolerance, used with :func:`numpy.isclose`.
@@ -684,16 +700,18 @@ def dottest(forward, adjoint, fwd_size, adj_size, fwd_complex=False,
         """Create random data of size and dtype of <size>."""
         out = rng.standard_normal(int(np.real(size)))
         if iscomplex:
-            out = out + 1j*rng.standard_normal(out.size)
+            out = out + 1j * rng.standard_normal(out.size)
         return out
 
     # Create random vectors u and v.
-    u = random(fwd_size, fwd_complex)
-    v = random(adj_size, adj_complex)
+    u = random(np.product(fwd_shape), fwd_complex)
+    v = random(np.product(adj_shape), adj_complex)
 
     # Carry out dot product test.
-    lhs = np.vdot(v, forward(u))  # lhs := v^H * (fwd  * u)
-    rhs = np.vdot(adjoint(v), u)  # rhs := (adj * v)^H * u
+    fwdu = forward(u.reshape(fwd_shape)).ravel(fwd_order)  # fwd * u
+    adjv = adjoint(v.reshape(adj_shape)).ravel(adj_order)  # adj * v
+    lhs = np.vdot(v, fwdu)  # lhs := v^H * (fwd  * u)
+    rhs = np.vdot(adjv, u)  # rhs := (adj * v)^H * u
 
     # Check if they are the same.
     passed = np.isclose(rhs, lhs, rtol, atol)
--- a/discretize/tests.py
+++ b/discretize/tests.py
@@ -22,7 +22,7 @@ Functions
   rosenbrock
   get_quadratic
   setup_mesh
-  dottest
+  assert_isadjoint
 """
 
 import numpy as np
@@ -638,20 +638,17 @@ def get_quadratic(A, b, c=0):
 
 
 # TESTS
-def dottest(
+def assert_isadjoint(
     forward,
     adjoint,
-    fwd_shape,
-    adj_shape,
-    fwd_complex=False,
-    adj_complex=False,
-    fwd_order="C",
-    adj_order="C",
+    shape_in,
+    shape_out,
+    complex_in=False,
+    complex_out=False,
     clinear=True,
     rtol=1e-6,
     atol=0.0,
-    raise_error=False,
-    verb=False,
+    assert_error=True,
 ):
     r"""Dot product test for the forward operator and its adjoint operator.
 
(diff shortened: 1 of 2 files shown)


discretize/tests.py
```diff
@@ -633,6 +637,110 @@ def Quadratic(x, return_g=True, return_H=True):
...
+    adj_v = adjoint(v.reshape(adj_shape, order=adj_order)).ravel(adj_order)
+    if clinear:
+        lhs = np.vdot(v, fwd_u)  # lhs := v^H * (fwd * u)
+        rhs = np.vdot(adj_v, u)  # rhs := (adj * v)^H * u
+    else:
+        lhs = np.dot(v.real, fwd_u.real) + np.dot(v.imag, fwd_u.imag)
+        rhs = np.dot(adj_v.real, u.real) + np.dot(adj_v.imag, u.imag)
+
+    # Check if they are the same.
+    passed = np.isclose(rhs, lhs, rtol, atol)
```
jcapriot, line 725: probably just use `np.testing.assert_allclose(rhs, lhs, rtol, atol)`

This function works well within testing environments, and will automatically raise correct errors, with much more intuitive error messages.

You can also pass it an `err_msg` kwarg to print when it fails as well. So maybe something like "Dot test failed".

discretize/tests.py
```diff
@@ -633,6 +637,110 @@ def Quadratic(x, return_g=True, return_H=True):
...
+
+    # Verbosity and error.
+    if (not passed and raise_error) or verb:
```
jcapriot, line 728: And then here, just remove the rest of the function.  Which, I believe would remove the use of the `verb` parameter.

tests/base/test_tests.py
```diff
@@ -0,0 +1,67 @@
...
+        h = [[11, 20.5, 33], [222, 111, 333], np.ones(4) * 55.5]
+        mesh1 = discretize.TensorMesh(h, "CCC")
+
+        hx = np.ones(20) * 20
+        mesh2 = discretize.TensorMesh([hx, hx, hx], "CCC")
+
+        P = discretize.utils.volume_average(mesh1, mesh2)
+
+        # return
+        assert dottest(
```
jcapriot, line 21: if dottest raises the proper assertions (when switching to `np.testing.assert_allclose`, then we do not need to `assert` the dot test.

discretize/tests.py
```diff
@@ -633,6 +637,110 @@ def Quadratic(x, return_g=True, return_H=True):
     return Quadratic
 
 
+# TESTS
+def dottest(
```
jcapriot, line 641: how about we rename this function to `assert_isadjoint` ? Maybe follow the `np.testing` naming scheme for it.

jcapriot (commented):
Left some review comments here

prisae:
Thanks @jcapriot for the feedback, I will have to think about them. I thought about it more like a user-function, that can be used in general. Your comments go more towards it being a testing utility, using the testing convention/things. I like it, makes probably sense!

prisae:
Just to check @jcapriot - the `OrderTest` and the `check_derivative` in the same file are verbose, and do not `assert`, but check. I thought of the `dottest` to be a similar routine, similarly used. Are the plans to change the `OrderTest` & `check_derivative` to also more resemble `np.testing/assertion` routines?

discretize/tests.py
```diff
@@ -633,6 +637,110 @@ def Quadratic(x, return_g=True, return_H=True):
...
+        if iscomplex:
+            out = out + 1j * rng.standard_normal(out.size)
+        return out
+
+    # Create random vectors u and v.
+    u = random(np.product(fwd_shape), fwd_complex)
+    v = random(np.product(adj_shape), adj_complex)
+
+    # Carry out dot product test.
+    fwd_u = forward(u.reshape(fwd_shape, order=fwd_order)).ravel(fwd_order)
```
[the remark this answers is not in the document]
prisae: This won't work if the `forward` and the `adjoint` take different shapes.

discretize/tests.py
prisae: Goes to the main discussion - should it be a check or an assertion.

discretize/tests.py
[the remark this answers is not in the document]
jcapriot: As long as the out put of the forward matches adj_shape, and the output of adj match’s forward shape. It will work. The two functions do not need to accept the same shaped argument.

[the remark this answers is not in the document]
prisae: This is/should be true indeed. I had a test example where it failed, but probably my example was bonkers! I'll look at it again.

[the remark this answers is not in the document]
jcapriot: If the functions returned the appropriate shaped arrays, you could even flatten them out irrespective of their internal ordering, as long as they're both flattened out with the same ordering.

jcapriot:
I do not like the way `check_derivative` and `OrderTest` are currently used for tests. When they fail, the only output you get is "AssertionError: False is not True" which is horribly nondescript. To find out why they actually failed, you need to search through the entire output of `pytest -s -v`.

If all you want for development testing environment is for it to write out the result of wTJv and (vJ)Tw, then simply add a print statement comparing the two before `np.testing.assert_allclose`

jcapriot:
Also the names `fwd_shape` and `adj_shape` are slightly too nondescript in my opinion. Do they mean "what is the shape of the matrix represented by this function?" because how they are used answers the question: "What is the output shape of this function?" Maybe change their names to `fwd_in_shape` and `fwd_out_shape`? Then by definition the adjoint would accept a `fwd_out_shape` and return a `fwd_in_shape`.

discretize/tests.py
prisae: Implemented.

discretize/tests.py
prisae: I removed the `verb` parameter, and renamed `raise_error` to `assert_error` (True by default). As such, I kept the "verbose" part so it could either be used as an assertion in tests, or as whatever the users want it to returning the boolean test result.

discretize/tests.py
prisae: Done, good idea.

tests/base/test_tests.py
prisae: Done.

prisae:
Thanks for the clarification regarding `OrderTest` and `check_derivative`, that makes things clearer. I implemented most of your suggestions. I further simplified `fwd_in_shape and `fwd_out_shape` to simply `shape_in` and `shape_out`; accordingly, I renamed `fwd_complex` and `adj_complex` to `complex_in` and `complex_out`.

Regarding assertion and verbosity: Have a look, I sort of made a compromise. By default, it acts now as a regular assertion test: `assert_isajdoint()` will be silent if passed, or raise the proper error if failed, using `np.testing.assert_allclose()`. However, if `assert_error` is set to False, it will instead print a message and return the test result as boolean. So it can still be used in other ways than pure testing. Would that be OK?

prisae:
The alternative would be to call it `shape_u`/`shape_v` and `complex_u`/`complex_v` to follow the docstring and code, instead of `_in`/`_out`.

discretize/tests.py
```diff
@@ -633,6 +637,121 @@ def Quadratic(x, return_g=True, return_H=True):
...
+
+    .. math::
+
+        \mathbf{v}^H ( \mathbf{F} \mathbf{u} ) =
+        ( \mathbf{F}^H \mathbf{v} )^H \mathbf{u}
+
+
+    Parameters
+    ----------
+    forward : function
```
jcapriot, line 665: Forward and adjoint should be `callable` not function.

discretize/tests.py
```diff
@@ -633,6 +637,121 @@ def Quadratic(x, return_g=True, return_H=True):
...
+
+    assert_error : bool, default: True
+        By default this test is an assertion (silent if passed, raising an
+        assertion error if failed). If set to False, the result of the test is
+        returned as boolean and a message is printed.
+
+
+    Returns
+    -------
+    passed : bool, returned if assert_error=False
```
jcapriot, line 704: this line should only have the dtype with it, put any descriptions on the next line

```
passed : bool, optional
    Result of the dot product test. returned if `assert_error` is False.
```

jcapriot (commented):
Looks good to me @prisae, just a few minor changes on the docstring, then good to go.

discretize/tests.py
prisae: Yes! I knew `function` was wrong, but I couldn't remember what it should be. Thanks.

discretize/tests.py
prisae: OK

PR #273, opened 2022-03-01
Move from bumpversion to setuptools_scm

prisae:
Draft implementation, still a few crucial bits missing.

- [ ] I have to look at the conda-forge recipe, I think it needs updating.
- [ ] How are the docs built? The way it is now requires that discretize is built in order t generate the docs.
- [ ] The main open thing is `MANIFEST.in` - it is very counter-intuitive. I think normally you use it to *include* things for PyPi. However, using `setuptools_scm` it includes everything by default, and you have to use `MANIFEST.in` to *exclude* stuff.

Also related, I will look into it to set up the GitHub Action, so the workflow would be:
- [ ] How is it deployed? Creating a release on GitHub will trigger the action, run all tests, and then deploy it automatically to PyPi.
- [ ] Every merge to main deploys to TestPyPi, so there should be no surprises on releases.

Probably other odd things when switching to setuptools_scm that I forgot, but hopefully will resurface in the next few days.


jcapriot:
I'm happy with where this is currently at @prisae, I'm going to merge this in.

jcapriot (approved):

prisae:
Awesome @jcapriot, thanks!

PR #283, opened 2022-07-22
add python 3.10 to testing suite

lheagy:


jcapriot:
Awesome, can you add it to the azure-pipelines.yml file in the base directory, where it builds the windows wheels

azure-pipelines.yml
```diff
@@ -108,6 +108,8 @@ stages:
             python.version: '3.8'
           win-Python39:
             python.version: '3.9'
+          win-Python310:
+            python.version: '3.10'
```
prisae, line 112: 7 lines above, there is still Python version 3.6 - good moment to remove?

jcapriot: probably a good idea, in that we don't even test it against 3.6 anymore

PR #284, opened 2022-07-23
Improve load time

prisae:
# Improve loadtime

## Motivation

`matplotlib`, `vtk`, and `omf` are soft dependencies. However, if they are installed they are, currently, loaded whether or not the mixins are used. This adds significantly to the load overhead (>1s).

This might seem irrelevant. However, for packages that use discretize and provide a command-line interface, it means their application takes an unnecessary second extra to start up.

## Implementation

This commit
- makes them load lazily, only loading if they are actually used;
- implements a simple load-test, ensuring that the load-time of discretize is below a defined threshold (currently 0.5s).

## Other thoughts

The remaining big portions of import-time are:
1. `numpy`: Any package using discretize is most likely also using numpy, so that is a not an issue.
2. Some submodules of `scipy`: However, SciPy will, starting with `v1.9.0`, lazy-load its submodules, so this will fix itself down the road. We will just have to change how we access it (for the interested, see SciPy-PR #15230).

## Screenshots

```
python -Ximporttime -c "import discretize" 2> discretize.log
tuna discretize.log
```

Assuming you have `matplotlib`, `vtk`, and `omf` installed, this is the current load time: Loading discretize takes almost 1.5s, of which roughly 75% are spent on importing `matplotlib`, `vtk`, and `omf`, even if they are never used.

main

This is how it looks with this PR: The load time is down from ~1.5s to ~0.33s.
improved

### DIFF
--- a/discretize/mixins/__init__.py
+++ b/discretize/mixins/__init__.py
@@ -27,33 +27,28 @@ Other Optional Classes
 
   Slicer
 """
+import importlib
 from .mesh_io import TensorMeshIO, TreeMeshIO, SimplexMeshIO
 
 AVAILABLE_MIXIN_CLASSES = []
 SIMPLEX_MIXIN_CLASSES = []
 
-try:
+if importlib.util.find_spec("vtk"):
     from .vtk_mod import InterfaceVTK
 
     AVAILABLE_MIXIN_CLASSES.append(InterfaceVTK)
-except ImportError as err:
-    pass
 
-try:
+if importlib.util.find_spec("omf"):
     from .omf_mod import InterfaceOMF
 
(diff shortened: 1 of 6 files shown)


jcapriot:
I'm not sure what is going on with the windows build errors, but it seems to be affecting all of the PRs, I'll check into it as soon as I can.

prisae:
Googling the corresponding windows error leads to this potential reason of the Windows fail:
https://stackoverflow.com/a/65826036

jcapriot:
I like the simplicity of this currently, and looks like everything is covered well by the tests. I'll get this in.

jcapriot (approved):

PR #285, opened 2022-08-02
zeros_outside

prisae:
Calling `plot_slice` with a `TensorMesh` currently results in the following warning:

```
/path/to/conda/env/lib/python3.9/site-packages/discretize/base/base_tensor_mesh.py:785: FutureWarning: The zerosOutside keyword argument has been deprecated, please use zeros_outside. This will be removed in discretize 1.0.0
  warnings.warn(
```

This is an internal problem, and should not happen.

### DIFF
--- a/discretize/mixins/mpl_mod.py
+++ b/discretize/mixins/mpl_mod.py
@@ -1203,7 +1203,7 @@ class InterfaceMPL(object):
             U, V = self.reshape(v.reshape((self.nC, -1), order="F"), "CC", "CC", "M")
 
             tMi = self.__class__(h=[hx, hy], origin=np.r_[origin_x, origin_y])
-            P = self.get_interpolation_matrix(tMi.gridCC, "CC", zerosOutside=True)
+            P = self.get_interpolation_matrix(tMi.gridCC, "CC", zeros_outside=True)
 
             Ui = tMi.reshape(P * mkvc(U), "CC", "CC", "M")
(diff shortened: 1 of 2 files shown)


prisae:
This started to appear in https://github.com/simpeg/discretize/commit/dcbfc4eddfaee5fa0050d42495bd83717a74164c, which is OK, basically changing the DeprecationWarning to a FutureWarning as planned.

However, when doing this we have to make sure that we do not have any internal warnings.

jcapriot (approved):
LGTM

prisae:
There are potentially many other warnings that might surface, I don't know and I haven't tested! This was just the one that popped up in the emg3d-gallery...

jcapriot:
I did a search for `zerosOutside` In the code base and this was the only spot that still had it.

prisae:
Yes, zerosOutside should be fine. But having had a quick glance at the test logs I had the impression there were many (internal) warnings for other updated parameters. But doesn't really matter, we fix them as they surface.

jcapriot:
Most of the tests are left alone specifically because I want them throwing the warnings to hit those lines and ensure backward compatibility. That being said I will start a `1.0.0` branch that will remove all of the pep8 conversion leftovers.

prisae:
Fair enough.

For that I find the context manager of pytest great
```
with pytest.warns(SomeWarning...):
    do the test
```
It ensures the warning is raised, but it is not a warning in the test log, so one can aim at a zero warning tests. (You probably know all of that, just in case.)

PR #288, opened 2022-08-11
Allow np.int_

prisae:
We should allow numpy int's in the plotting, as matplotlib handels them just fine.

We could also clean up some of the code on the way, such as `int(np.argmi(...))` could be simplified with `np.argmin()`, as it returns a `np.int_`.

Currently it simply fails (for no good reasons). -- Thoughts? (Maybe we should not check at all, and let matplotlib fail if it doesn't work, duck typing.)

### DIFF
--- a/discretize/mixins/mpl_mod.py
+++ b/discretize/mixins/mpl_mod.py
@@ -2089,15 +2089,14 @@ class InterfaceMPL(object):
         szSliceDim = len(self.h[normalInd])
         if ind is None:
             ind = szSliceDim // 2
+        if not isinstance(ind, (np.integer, int)):
+            raise ValueError("ind must be an integer")
 
         cc_tensor = [None, None, None]
         for i in range(3):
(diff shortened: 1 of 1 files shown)


jcapriot:
Should we check instead for `np.integer`? That way it catches other types of integers as well, i.e. `np.int32`, `np.int8`

prisae:
Good call, I'll change it.

Question remains, do we need to / should we check at all?

jcapriot:
It doesn't hurt? I think the intended purpose of checking it here, is to give a slightly more customized error message than the one caused by indexing an array with a non-integer.

prisae:
That is true. The other option would be to simply cast it `int(ind)`, but it is fine as it is. Still, including `(np.integer, int)` makes sense I think.

prisae:
I just pushed another commit, one of the two integer checks was actually after `ind` has been used.

jcapriot:
looks good, thanks again

PR #297, opened 2022-11-23
Add new refine surface, bounding box, and point refine methods.

jcapriot:
This is intended to replace the functionality of the `refine_tree_xyz` utility, which will be marked as deprecated.
The new functions have slightly more intuitive inputs (in my opinion). They use the `level` argument to explicitly refer to the octree level, and `padding_cells_by_level` to refer to the amount of padding around the object.

The previous function split the padding descriptions into the `octree_levels` argument and the `octree_levels_padding`. The method of refining at specific levels is not intuitive and it was implied only by the length of the array passed to `octree_levels`, and not there values, as the name might imply.

~There is still missing tests~, but just wanted to get these out there for feedback on the inputs and descriptions in the documentation of these functions. The new `surface` function is not necessarily any quicker than the previous, but it is more well behaved. Although it's cost rises faster with the number of triangles in the triangulation.

### DIFF
--- a/discretize/_extensions/tree_ext.pyx
+++ b/discretize/_extensions/tree_ext.pyx
@@ -857,6 +857,10 @@ cdef class _TreeMesh:
             Whether to balance cells diagonally in the refinement, `None` implies using
             the same setting used to instantiate the TreeMesh`.
 
+        See Also
+        --------
+        refine_surface
+
         Examples
         --------
         We create a simple mesh and refine the TreeMesh such that all cells that
@@ -865,23 +869,25 @@ cdef class _TreeMesh:
         >>> import discretize
         >>> import matplotlib.pyplot as plt
         >>> import matplotlib.patches as patches
-        >>> tree_mesh = discretize.TreeMesh([32, 32])
+        >>> tree_mesh = discretize.TreeMesh([32, 32, 32])
         >>> tree_mesh.max_level
         5
 
-        Next we define the points along the line and the level we want to refine to,
-        and refine the mesh.
+        Next we define the bottom points of the prism, its heights, and the level we
+        want to refine to, then refine the mesh.
 
-        >>> triangle = [[0.14, 0.31], [0.32, 0.96], [0.23, 0.87]]
+        >>> triangle = [[0.14, 0.31, 0.21], [0.32, 0.96, 0.34], [0.87, 0.23, 0.12]]
+        >>> height = 0.35
         >>> levels = 5
-        >>> tree_mesh.refine_triangle(triangle, levels)
+        >>> mesh.refine_vertical_trianglular_prism(triangle, height, levels)
 
-        Now lets look at the mesh, and overlay the line on it to ensure it refined
-        where we wanted it to.
+        Now lets look at the mesh.
 
-        >>> ax = tree_mesh.plot_grid()
-        >>> tri = patches.Polygon(triangle, fill=False)
-        >>> ax.add_patch(tri)
+        >>> v = mesh.cell_levels_by_index(np.arange(mesh.n_cells))
+        >>> fig, axs = plt.subplots(1, 3, figsize=(12,4))
+        >>> mesh.plot_slice(v, ax=axs[0], normal='x', grid=True, clim=[2, 5])
+        >>> mesh.plot_slice(v, ax=axs[1], normal='y', grid=True, clim=[2, 5])
+        >>> mesh.plot_slice(v, ax=axs[2], normal='z', grid=True, clim=[2, 5])
         >>> plt.show()
 
         """
--- a/discretize/_extensions/tree_ext.pyx
+++ b/discretize/_extensions/tree_ext.pyx
@@ -2242,7 +2242,6 @@ cdef class _TreeMesh:
         of the x and y-boundary cells.
 
         >>> from discretize import TreeMesh
-        >>> from discretize.utils import refine_tree_xyz
         >>> import numpy as np
         >>> import matplotlib.pyplot as plt
 
@@ -2341,7 +2340,6 @@ cdef class _TreeMesh:
         of the x and y-boundary faces.
 
         >>> from discretize import TreeMesh
-        >>> from discretize.utils import refine_tree_xyz
         >>> import numpy as np
         >>> import matplotlib.pyplot as plt
 
--- a/discretize/_extensions/tree_ext.pyx
+++ b/discretize/_extensions/tree_ext.pyx
@@ -521,9 +521,9 @@ cdef class _TreeMesh:
         ----------
         points : (N, dim) array_like
             The centers of the refinement balls
-        radii : (N) array_like
+        radii : float or (N) array_like of float
             A 1D array defining the radius for each ball
-        levels : (N) array_like of int
+        levels : int or (N) array_like of int
             A 1D array defining the maximum refinement level for each ball
         finalize : bool, optional
             Whether to finalize after refining
@@ -566,12 +566,19 @@ cdef class _TreeMesh:
         if points.shape[1] != self.dim:
             raise ValueError(f"points array must be (N, {self.dim})")
         cdef double[:, :] cs = points
-        cdef double[:] rs = np.require(np.atleast_1d(radii), dtype=np.float64,
+        radii = np.require(np.atleast_1d(radii), dtype=np.float64,
                                        requirements='C')
+        if radii.shape[0] == 1:
+           radii = np.full(points.shape[0], radii[0], dtype=np.float64)
+        cdef double[:] rs = radii
         if points.shape[0] != rs.shape[0]:
             raise ValueError("radii length must match the points array's first dimension")
-        cdef int[:] ls = np.require(np.atleast_1d(levels), dtype=np.int32,
+
+        levels = np.require(np.atleast_1d(levels), dtype=np.int32,
                                     requirements='C')
+        if levels.shape[0] == 1:
+            levels = np.full(points.shape[0], levels[0], dtype=np.int32)
+        cdef int[:] ls = levels
         if points.shape[0] != ls.shape[0]:
             raise ValueError("level length must match the points array's first dimension")
 
@@ -603,11 +610,11 @@ cdef class _TreeMesh:
             The minimum location of the boxes
         x1s : (N, dim) array_like
             The maximum location of the boxes
-        levels : (N) array_like of int
+        levels : int or (N) array_like of int
             The level to refine intersecting cells to
         finalize : bool, optional
             Whether to finalize after refining
-        diagonal_balance : bool or None, optional
+        diagonal_balance : None or bool, optional
             Whether to balance cells diagonally in the refinement, `None` implies using
             the same setting used to instantiate the TreeMesh`.
 
@@ -640,7 +647,6 @@ cdef class _TreeMesh:
         >>> rect = patches.Rectangle([0.8, 0.8], 0.1, 0.2, facecolor='none', edgecolor='k', linewidth=3)
         >>> ax.add_patch(rect)
         >>> plt.show()
-
         """
         x0s = np.require(np.atleast_2d(x0s), dtype=np.float64,
                             requirements='C')
@@ -654,8 +660,11 @@ cdef class _TreeMesh:
             raise ValueError(f"x0s and x1s must have the same length")
         cdef double[:, :] x0 = x0s
         cdef double[:, :] x1 = x1s
-        cdef int[:] ls = np.require(np.atleast_1d(levels), dtype=np.int32,
+        levels = np.require(np.atleast_1d(levels), dtype=np.int32,
                                     requirements='C')
+        if levels.shape[0] == 1:
+            levels = np.full(x0.shape[0], levels[0], dtype=np.int32)
+        cdef int[:] ls = levels
         if x0.shape[0] != ls.shape[0]:
             raise ValueError("level length must match the points array's first dimension")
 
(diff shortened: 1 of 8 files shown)


jcapriot:
I'm thinking it might be a faster option to refine the triangulated surface (and not pass it as the simplexes), and shift it up and down for each padding level. The triangle refine is quicker than the tetrahedron refine as there are half as many edges to check and a third of the faces...

discretize/utils/mesh_utils.py
```diff
@@ -693,8 +710,20 @@ def refine_tree_xyz(
...
     # Trigger different refine methods
     if method.lower() == "radial":
+        # padding = octree_levels[non_zeros]
+        # mesh.refine_points(xyz, levels, padding, finalize=finalize,)
+        warnings.warn(
+            "The radial option is deprecated as of `0.9.0` please update your code to "
+            "use the `TreeMesh.refine_points` functionality. It will be removed in a "
+            "future version of discretize.",
+            DeprecationWarning,
+        )
```
domfournier, line 726: Strangely the warning doesn't print on my end... even though I am sure I am getting there

image

discretize/_extensions/tree_ext.pyx
```diff
@@ -521,9 +521,9 @@ cdef class _TreeMesh:
         ----------
```
domfournier, line 521: In case you are not seeing on your side - I am getting a lot of warnings when compiling the code
image

domfournier:
> ~I'm thinking it might be a faster option to refine the triangulated surface (and not pass it as the simplexes), and shift it up and down for each padding level. The triangle refine is quicker than the tetrahedron refine as there are half as many edges to check and a third of the faces, not to mention for each triangle of the surface, it takes 3 tetrahedron to represent the triangular prism column that is currently represented.~
> 
> Turns out it is about the same

Confirmed, almost the same for 200k triangles and ~500k cells mesh.
image

image

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+        that surrounds those points. The arbitrary points we use for this example are
+        uniformly scattered between [3/8, 5/8] in the first and second dimension.
+
+        >>> import discretize
+        >>> import matplotlib.pyplot as plt
+        >>> import matplotlib.patches as patches
+        >>> mesh = discretize.TreeMesh([32, 32])
+        >>> points = np.random.rand(20, 2) * 0.25 + 3/8
+
+        Now we want to refine to the maximum level, with a no padding the in `x`
```
domfournier, line 404:
```suggestion
        Now we want to refine to the maximum level, with a no padding in the `x`
```

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+        padding cells at each dimension.
+
+        Parameters
+        ----------
+        points : (N, dim) array_like
+            The bounding box will be the maximum and minimum extent of these points
+        level : (n_level) array_like of int
+            The level of the treemesh to refine the bounding box to. If an array it will
+            refine the box to each level (mostly useful in combination with `padding_cells_by_level`)
+            Negative values index backwards, (e.g. `-1` is `max_level`).
```
domfournier, line 380: This is a bit annoying to have to count backward - will result in a lot of explanations. Can we not do the translation for the user?

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+        It also optionally pads the bounding box at each level based on the number of
+        padding cells at each dimension.
+
+        Parameters
+        ----------
+        points : (N, dim) array_like
+            The bounding box will be the maximum and minimum extent of these points
+        level : (n_level) array_like of int
+            The level of the treemesh to refine the bounding box to. If an array it will
+            refine the box to each level (mostly useful in combination with `padding_cells_by_level`)
```
domfournier, line 379: I am not getting this part - Feels like the level as `array` only matters if you apply padding_cells. Should we not restrict `level: int` for this one?

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+            Negative values index backwards, (e.g. `-1` is `max_level`).
+        padding_cells_by_level : None or (n_level) iterable, optional
+            The number of cells to pad the bounding box at each level of refinement.
+            Each entry may be either a single number, which means to pad equally in each
+            dimension, or it may be an (`dim`) `array_like` for variable padding along
+            each dimension. None implies no padding.
+        finalize : bool, optional
+            Whether to finalize the mesh after the call.
+        diagonal_balance : None or bool, optional
+            Whether to balance cells diagonally in the refinement, `None` implies using
```
domfournier, line 389:
```suggestion
            Whether to balance cells diagonally in the refinement. `None` implies using
```

discretize/tree_mesh.py
domfournier: ok I get it now - you are trying not to impose order in the values of levels. Feels like we are increasing the complexity of input arguments for very little benefit. I would propose we revert to combining `level` and `padding_cells_by_level` into a single `list[int]`.

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+        >>> mesh.plot_slice(v, ax=axs[1], normal='y', grid=True, clim=[2, 5])
+        >>> mesh.plot_slice(v, ax=axs[2], normal='z', grid=True, clim=[2, 5])
+        >>> plt.show()
+
+        """
+        level = np.atleast_1d(level)
+        # pad based on the number of cells at each level
+        if padding_cells_by_level is None:
+            padding_cells_by_level = np.zeros_like(level)
+        padding_cells_by_level = np.asarray(padding_cells_by_level, dtype=object)
```
domfournier, line 607: What's the `dtype=object` there for?

discretize/utils/mesh_utils.py
```diff
@@ -823,6 +863,18 @@ def refine_tree_xyz(
...
+            "The box option is deprecated as of `0.9.0` please update your code to "
+            "use the `TreeMesh.refine_bounding_box` functionality. It will be removed in a "
+            "future version of discretize.",
+            DeprecationWarning,
+        )
+        # padding = np.zeros((len(octree_levels), mesh.dim))
+        # padding[:, -1] = np.maximum(octree_levels - 1, 0)
+        # padding[:, :-1] = octree_levels_padding[:, None]
+        # padding = padding[non_zeros]
+        # mesh.refine_bounding_box(xyz, levels, padding, finalize=finalize)
```
domfournier, line 876: Same here...

discretize/utils/mesh_utils.py
```diff
@@ -712,6 +741,17 @@ def refine_tree_xyz(
...
+            "The surface option is deprecated as of `0.9.0` please update your code to "
+            "use the `TreeMesh.refine_surface` functionality. It will be removed in a "
+            "future version of discretize.",
+            DeprecationWarning,
+        )
+        # padding = np.zeros((len(octree_levels), mesh.dim))
+        # padding[:, -1] = np.maximum(octree_levels - 1, 0)
+        # padding[:, :-1] = octree_levels_padding[:, None]
+        # padding = padding[non_zeros]
+        # mesh.refine_surface(xyz, levels, padding, finalize=finalize, pad_down=True, pad_up=False)
```
domfournier, line 754: Why not re-routing to the new method now? These lines don't run as is BTW

discretize/tree_mesh.py
```diff
@@ -359,6 +360,316 @@ def origin(self, value):
...
+        padding_cells_by_level = padding_cells_by_level[sorted]
+
+        if self.dim == 2:
+            xyz = np.asarray(xyz)
+            sorted = np.argsort(xyz[:, 0])
+            xyz = xyz[sorted]
+            n_ps = len(xyz)
+            inds = np.arange(n_ps)
+            simps1 = np.c_[inds[:-1], inds[1:], inds[:-1]] + [0, 0, n_ps]
+            simps2 = np.c_[inds[:-1], inds[1:], inds[1:]] + [n_ps, n_ps, 0]
```
domfournier, line 624: I don't think we can assume that the order of the points will make nice simplices. We should probably ask for the triangulation instead...

domfournier:
@jcapriot Went through it with some testing. Two sticky things for me:
1- The definition of levels counting backward and padding cells seems odd an not easy to follow
2- Surface refinement should require arguments for the vertices AND simplices. It is also slower than the old method, although the 3D generalization is nice.

jcapriot:
1.) In my opinion the current implementation on `refine_xyz` of inferring the level by the length of an input array is not intuitive. This method allows you to explicitly control which level you want to refine to. Regarding the negative input values, `-1` and `mesh.max_level` both refer to the same thing. It's used the same way you can index the last value of an array in python with negative levels. All of the `mesh.refine_` functions on tree mesh work this way.
2.) Can you give me those input meshes so I can play around with the internals a bit?

discretize/tree_mesh.py
jcapriot: In 2D the input is sorted first to ensure they're nice. In 3D it currently supports both methods of `(vertices, simplexes)` or just `xyz`

discretize/utils/mesh_utils.py
jcapriot: These are here to mostly just remind ourselves of the translations between the two methods, as their calling conventions are slightly different, and the surface refine behaves slightly differently around the edges.

discretize/tree_mesh.py
jcapriot: so you could pass an array with different length elements, (i.e. `padding_cells_by_level=[1 , [0, 0, 1]]`)

discretize/tree_mesh.py
jcapriot: I would rather leave the refinement level as an explicit input argument that is not inferred by the length of an input list.

discretize/tree_mesh.py
jcapriot: You don't have to count backwards, you could also supply the refinement level as well. The negative indexing works just like array indexing in python.

domfournier:
I like the dictionary idea. I still think that working with negative indexing is cumbersome as users never care for index 0,  always the max level then backward. This is just internal mechanics, we could make an effort to make it more intuitive.

discretize/tree_mesh.py
domfournier: I hear you, but still not very convenient.

discretize/tree_mesh.py
domfournier: Let s do the dictionary. More obvious that the two arguments depend on each other. Otherwise there is weird hidden dependency IMO

dccowan: It seems like the basic concept is that:
- we have a single box
- everywhere within that box will be discretize to the finest level provided (which means *level* makes more sense as an *int*)
- and *padding_cells_by_level* is how many layers of the current level of cells before cell size in increased

For *padding_cells_by_level*, I would personally like the have the following options:
- *int*: where the same number of padding cells is provided for every level
- *array_like* of *int*: since discretization below *level* is not done, we don't want define values in this argument that aren't used. If it is an array_like, the first entry provides the padding for refinement at *level* and we go from.

dccowan:
**Comment on efficacy for end users:** I feel like most people will want to set the finest desired level of discretization, then set the number of padding cells at level as things are more coarse. So having the *level* argument be *int* would be the easiest implementation. In addition, I think new users will naturally be inclined to set the finest discretization by having *level*= 1 instead of *level* = self.max_level. 

**Comment on level as an array_like:** If *padding_cells_per_level* is array_like, the function already knows what values correspond to which level even if *level* is *int*. Can someone provide a commonly encountered situation where *level* must be array_like in order to make the mesh they want?

jcapriot:
> **Comment on efficacy for end users:** I feel like most people will want to set the finest desired level of discretization, then set the number of padding cells at level as things are more coarse. So having the _level_ argument be _int_ would be the easiest implementation. In addition, I think new users will naturally be inclined to set the finest discretization by having _level_= 1 instead of _level_ = self.max_level.

I'm pretty hard against referring to `level` as opposite to how it is used everywhere else in the codebase just for these functions. Reversing the input by starting the finest cell level as 0/1 would lead to more confusion in my opinion, especially if you look at the mesh afterwards and see a 100,000 cells at the level 6. The negative indexing is meant to make it easier to refer to the finest level (instead of writing `mesh.max_level` or `mesh.max_level - 1`).


> I like the dictionary idea. I still think that working with negative indexing is cumbersome as users never care for index 0, always the max level then backward. This is just internal mechanics, we could make an effort to make it more intuitive.

What would make it more intuitive? I do want to force the user to be explicit about which level they want to refine to. (I've seen use cases before where users wanted to use the `refine_tree_xyz` function to not refine to the max level and were very confused about how to make it do that.)

discretize/tree_mesh.py
jcapriot: Hmm, that's also an interesting way of looking at describing the padding, and would disconnect the `level` and `padding_cells_by_level` argument. I'll ponder on that a little bit.

domfournier:
I think if a user creates a mesh and doesn't want to refine to its lowest, then they should use a bigger core cell size. But whatever, you have the final say.

jcapriot:
The example I recalled was they wanted to refine near the surface at the finest level, but then refine at a coarser level within a box.

domfournier:
That's fine, most users won't interact at that level anyway. Tree mesh creation is tough already, we can just create simple UIs and do the translation under the hood.

jcapriot:
What if `level` was a keyword argument with a default of `max_level`? I feel like that would be a good compromise, then also allow the padding to function like Devon’s description.

domfournier:
Yea that sounds good. So the `padding_cells_by_level: list[list | int]` would be relative to the `level: int` defaulted to max_level. If a user wants to start coarser, than they need to supply the true octree level index. Makes sense.

PR #322, opened 2023-05-29
Capture scroll

prisae:
The behaviour changed with the advent of ipympl (Lab; modern Notebook), I think. Currently the scroll over a figure makes the notebook keep scrolling, so it is hard to scroll over a slicer-image through the volume.

This PR implements the fix suggested by Matplotlib (PR 235).

### DIFF
--- a/discretize/mixins/mpl_mod.py
+++ b/discretize/mixins/mpl_mod.py
@@ -2580,6 +2580,8 @@ class Slicer(object):
         # Create subplots
         self.fig = plt.gcf()
         self.fig.subplots_adjust(wspace=0.075, hspace=0.1)
+        # To capture mouse scroll in notebooks (otherwise it is hard to slice through volume)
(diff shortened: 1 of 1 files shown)


jcapriot:
Looks good to me, I'm assuming you've tested this yourself and it works well on your end too

prisae:
However, the `fig.canvas` attribute exists always, as in other parts of the code we set
```
fig.canvas.mpl_connect("scroll_event", tracker.onscroll)
```
to listen to the scroll events. So setting
```
self.fig.canvas.capture_scroll = True
```
Should always be possible, and I cannot imagine that it has any negative side effects. It was created for this reason.

Let me know if you would prefer that I do a more thorough testing.

prisae:
As reference maybe also this:
https://matplotlib.org/ipympl/examples/full-example.html#comprehensive-example
(9th cell from the top)

PR #324, opened 2023-06-01
Add `total_nodes` method to TreeMeshes

santisoler:
The new method returns the whole set of nodes in the mesh, including non-hanging and hanging nodes. It comes useful when working together with `cell_nodes`.

### DIFF
--- a/discretize/tree_mesh.py
+++ b/discretize/tree_mesh.py
@@ -742,6 +742,21 @@ class TreeMesh(
             self.finalize()
 
     @property
+    def nodes_all(self):
+        """Gridded hanging and non-hanging nodes locations.
+
+        This property returns a numpy array of shape
+        ``(n_nodes + n_hanging_nodes, dim)`` containing gridded locations for
+        all hanging and non-hanging nodes in the mesh.
+
+        Returns
+        -------
+        (n_nodes + n_hanging_nodes, dim) numpy.ndarray of float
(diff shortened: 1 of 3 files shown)


jcapriot:
This should probably be called `total_nodes` to follow the naming convention already in the tree mesh: `nodes`, `hanging_nodes`, `n_nodes`, `n_hanging_nodes` and `n_total_nodes`.

santisoler:
I agree that it should follow the convention used on the other methods, so I think it's better to rename it.

Nevertheless, I think is nice to name things like `nodes_all`, `nodes_hanging`, `nodes_non_hanging` for easier discoverability: they'll be listed together in the docs, users can use tab completion to explore what properties and methods are available for `nodes...`. Just mentioning to keep it in mind for the future.

discretize/tree_mesh.py
```diff
@@ -741,6 +741,21 @@ def refine_surface(
...
+    def total_nodes(self):
+        """Gridded hanging and non-hanging nodes locations.
+
+        This property returns a numpy array of shape
+        ``(n_nodes + n_hanging_nodes, dim)`` containing gridded locations for
+        all hanging and non-hanging nodes in the mesh.
+
+        Returns
+        -------
+        (n_nodes + n_hanging_nodes, dim) numpy.ndarray of float
```
jcapriot, line 754: n_total_nodes

PR #325, opened 2023-06-02
Add new TensorCell class

santisoler:
The new `TensorCell` class represents a single cell in a TensorMesh, with `center`, `bounds`, `neighbors`, `nodes`, `edges` and `faces` properties. Add `__iter__` and `__getitem__` methods to `TensorMesh`. The latter allows to obtain a single cell in the mesh based on their indices. The former allows to iterate over every cell in the mesh. Add tests for the new class and methods.


discretize/tensor_mesh.py
```diff
@@ -159,6 +162,53 @@ def _repr_html_(self):
...
+            indices = ((i, j) for j in range(ny) for i in range(nx))
+        elif self.dim == 3:
+            nx, ny, nz = self.shape_cells
+            indices = (
+                (i, j, k) for k in range(nz) for j in range(ny) for i in range(nx)
+            )
+        iterator = (self[index] for index in indices)
+        return iterator
+
+    def __getitem__(self, indices):
```
jcapriot, line 183: I think it should accept both linear indices, `[0, n_cells)` and tuples, `[ix, iy, iz]`. I would also rely on numpy's `ravel_multi_index` function reliably correctly convert from the tuple to the linear index.

santisoler: Done! By allowing `__getitem__` to accept integer indices, I could greatly simplify the `__iter__`.

jcapriot:
I'm trying to think if it would be good to just not allow a User to directly instantiate a TensorCell directly, We could do this by internally defining a `__new__` method and interacting with that internally, then throwing an exception in the class's `__init__`.

jcapriot:
Essentially it means you don't explicitly need to do all of the error checking for proper input. There's nothing wrong about creating the object on it's own, it's just that it does really have any tangible meaning without the Mesh class that it came from.

santisoler:
I agree that this new class doesn't introduce anything new if it's isolated from the mesh. That's why I think users won't be manually creating `TensorCell`s.

If I understand you correctly, you think it doesn't worth adding all the error raising on invalid inputs because users won't be creating this objects manually, so their proper creation should be covered by tests for the `TensorMesh.__getitem__()` method, right?

jcapriot:
> If I understand you correctly, you think it doesn't worth adding all the error raising on invalid inputs because users won't be creating this objects manually, so their proper creation should be covered by tests for the `TensorMesh.__getitem__()` method, right?

Yes, exactly. Unless we are concerned about ourselves programming this incorrectly to start with I don't think we necessarily need all the checks.

jcapriot:
Also, if the cell knows the shape of the mesh it came from, it can determine for itself the neighbor, node, face, etc. indices without passing it the mesh.

santisoler:
> Also, if the cell knows the shape of the mesh it came from, it can determine for itself the neighbor, node, face, etc. indices without passing it the mesh.

Sounds good, that's something I can add.

santisoler:
@jcapriot I think this is ready to be reviewed. Is there anything else we would like to add to the `TensorCell` class?

santisoler:
For some reason, Windows machines didn't like to check if a variable is an integer with 
```python
np.issubdtype(type(variable), int)
```
But using `np.integer` instead solved the issue: 
```python
np.issubdtype(type(variable), np.integer)
```

jcapriot:
> 

That makes sense, `int` is not technically a numpy `dtype`

jcapriot (commented):
Looks good, only thing left to do is to actually add it to the documentation pages.

We will likely need to re-name the `Tree Mesh Cells` section in `__init__.py` to just be `Mesh Cells` and update the text of that section accordingly (and add the `TensorCell`) to that list.

Also can you add a simple example to the docs of the `TensorCell` ?

PR #326, opened 2023-06-12
Face props mass matrices

dccowan:
Pull request for face properties mass matrix


jcapriot:
Did you try this with just the `average_edge_to_face` matrices? I think those should be all you need (similar to how the inner product matrices use `average_edge_to_cell` to create themselves.

discretize/base/base_tensor_mesh.py
```diff
@@ -996,6 +990,346 @@ def innerProductDeriv(v=None):
...
+                n_elements = sum([1 if x != 0 else 0 for x in shape]) - 1
+        else:
+            n_elements = self.dim - 1
+
+        # Isotropic case only
+        if model.size == self.nF:
+            Aprop = self.face_areas * mkvc(model)
+            if projection_type == 'E':
+                Av = getattr(self, "average_edge_to_face")
+                M = n_elements * sdiag(Av.T * Aprop)
```
lheagy, line 1053: I think we might need to be a bit careful here. When we take the transpose of this operator, 

```
mesh = discretize.TensorMesh([2, 1, 1])
A = mesh.average_edge_to_face.todense()
```

If we take the sum across all columns, they should sum to 1 (e.g. this is a True average) 
```
A.sum(1)
```
which it does. 

But, if we take the transpose, the columns don't sum to 1 
```
A.T.sum(1) 
```
image


I think what we should be doing is define 
```
Ae2f = mesh.average_edge_to_face.todense()
w = Ae2f.T.sum(1)
Af2e = discretize.utils.sdiag(1./w) * Ae2f.T
```

lheagy: nevermind! Thanks for straightening this out for me @jcapriot. Conceptually, we think about averaging the field to cell centers and then taking the product

discretize/base/base_mesh.py
```diff
@@ -1880,6 +1880,46 @@ def get_edge_inner_product(
...
+    def get_face_inner_product_surface(
+        self,
+        model=None,
+        invert_model=False,
+        invert_matrix=False,
+        do_fast=True,
+        **kwargs,
+    ):
+
+        raise NotImplementedError(
```
jcapriot, line 1906: You could actually implement this one for all mesh types at the same time, As the operation you need to do is just:

```
diag(face_areas * model)
```

and `face_areas` is defined for all mesh types.

discretize/base/base_mesh.py
```diff
@@ -1880,6 +1880,46 @@ def get_edge_inner_product(
...
+
+    def get_edge_inner_product_line(
+        self,
+        model=None,
+        invert_model=False,
+        invert_matrix=False,
+        do_fast=True,
+        **kwargs,
+    ):
+
```
jcapriot, line 1918: similar to above, you can do this for all mesh types right here:

```
diag(edge_lengths * model)
```

etc.

tests/base/test_tensor_innerproduct.py
```diff
@@ -162,6 +162,166 @@ def test_order6_faces_invert_model(self):
...
+        for ii, comp in enumerate(["x", "y", "z"]):
+            k = np.isclose(
+                eval("self.M.faces_{}".format(comp))[:, ii], 0.5
+            )  # x, y or z location for each plane
+            tau_ii = 1e-8 * eval(
+                "np.ones(self.M.nF{})".format(comp)
+            )  # effectively zeros but stable
+            tau_ii[k] = eval("call(tau_{}, self.M.faces_{}[k, :])".format(comp, comp))
+            tau[ii] = tau_ii
+        tau = np.hstack(tau)
```
santisoler, line 194: After what you commented on today's SimPEG meeting regarding flake complaining about unused `tau_x`, `tau_y` and `tau_z` (although they are being used in the `eval` calls), I was thinking that using `eval` could be avoided in this case by:
- using `getattr` for the attributes and methods of objects, and
- define the lambda functions as values of a dict, and then iterate over the dictionary.

For example:
```suggestion
        tau_funcs = {
            "x": lambda x, y, z: y * z + 1,  # x-face properties
            "y": lambda x, y, z: x * z + 2,  # y-face properties
            "z": lambda x, y, z: 3 + x * y,  # z-face properties
        }

        tau = []
        for ii, (comp, tau_func_i) in enumerate(tau_funcs.items()):
            m = getattr(self, "M")
            k = np.isclose(
                getattr(m, f"faces_{comp}")[:, ii], 0.5
            )  # x, y or z location for each plane
            tau_ii = 1e-8 * np.ones(
                getattr(m, f"nF{comp}")
            )  # effectively zeros but stable
            tau_ii[k] = call(tau_func_i, getattr(m, f"faces_{comp}")[k, :])
            tau.append(tau_ii)
        tau = np.hstack(tau)
```

This also prevents violating PEP8 regarding assignments of `lambda` functions (https://peps.python.org/pep-0008/#programming-recommendations).

tests/base/test_tensor_innerproduct.py
```diff
@@ -162,6 +162,166 @@ def test_order6_faces_invert_model(self):
...
+    using edgeInnerProducts and faceInnerProducts."""
+
+    meshTypes = ["uniformTensorMesh"]
+    meshDimension = 3
+    meshSizes = [16, 32]
+
+    def getError(self):
+        call = lambda fun, xyz: fun(xyz[:, 0], xyz[:, 1], xyz[:, 2])  # NOQA E731
+
+        ex = lambda x, y, z: x**2 + y * z  # NOQA E731
```
jcapriot, line 176: Error E731 is already ignored in all of discretize's tests. All of the ` # NOQA E731` in here are unnecessary.

tests/base/test_tensor_innerproduct.py
```diff
@@ -162,6 +162,166 @@ def test_order6_faces_invert_model(self):
...
+
+        tau = 3 * [None]
+        for ii, comp in enumerate(["x", "y", "z"]):
+            k = np.isclose(
+                eval("self.M.faces_{}".format(comp))[:, ii], 0.5
+            )  # x, y or z location for each plane
+            tau_ii = 1e-8 * eval(
+                "np.ones(self.M.nF{})".format(comp)
+            )  # effectively zeros but stable
+            tau_ii[k] = eval("call(tau_{}, self.M.faces_{}[k, :])".format(comp, comp))
```
jcapriot, line 192: I think you could effectively test this with only a single function for `tau` instead of defining three different ones for each cell face, and then having to use this very odd line to evaluate everything. You should try that and see if it also works.

discretize/base/base_mesh.py
```diff
@@ -1880,6 +1883,415 @@ def get_edge_inner_product(
...
+                M = sdiag(Aprop)
+            else:
+                raise Exception(
+                    "Unexpected shape of tensor: {}".format(model.shape),
+                    "Must be scalar or have length equal to total number of faces.",
+                )
+        except NotImplementedError:
+            raise NotImplementedError(
+                f"get_face_inner_product_surface not implemented for {type(self)}"
+            )
```
jcapriot, line 2142: There are a lot of things being tried in this statement here, and it's slightly unclear to someone reading the code what statements you're actually testing for exceptions. I'd recommend as follows:

```suggestion
        try:
            areas = self.face_areas
        except NotImplementedError:
            raise NotImplementedError(
                f"get_face_inner_product_surface not implemented for {type(self)}"
            )

        if model is None:
            model = np.ones(self.nF)

        if invert_model:
            model = 1.0 / model

        if is_scalar(model):
            model = model * np.ones(self.nF)

        # Isotropic case only
        if model.size != self.n_faces:
            raise ValueError(
                f"Unexpected shape of tensor: {model.shape}. "
                "Must be scalar or have length equal to total number of faces.",
            )

        Aprop = areas * mkvc(model)
        M = sdiag(Aprop)
```

This makes it fail faster if the mesh hasn't implemented a `face_areas` property yet (and it makes it clear in the code that that is what you are testing whether or not it is implemented). It also separates the check for the size of the model from constructing the matrix. (It does the check first then proceeds if it is all good.)

discretize/base/base_mesh.py
```diff
@@ -1880,6 +1883,415 @@ def get_edge_inner_product(
...
+            else:
+                raise Exception(
+                    "Unexpected shape of tensor: {}".format(model.shape),
+                    "Must be scalar or have length equal to total number of edges.",
+                )
+
+        except NotImplementedError:
+            raise NotImplementedError(
+                f"get_edge_inner_product_line not implemented for {type(self)}"
+            )
```
jcapriot, line 2288: I would replace these lines with basically the same thing as I put for the face innerproduct surface function.

discretize/base/base_mesh.py
```diff
@@ -2246,6 +2657,285 @@ def get_edge_inner_product_deriv(
...
+                        self.nE
+                    ),
+                )
+
+            L = sdiag(self.edge_lengths)
+
+        except NotImplementedError:
+            raise NotImplementedError(
+                f"get_edge_inner_product_line_deriv not implemented for {type(self)}"
+            )
```
jcapriot, line 2881: Similar comments to above. Move the raised exceptions closer to what they are testing.

discretize/base/base_mesh.py
```diff
@@ -2246,6 +2657,285 @@ def get_edge_inner_product_deriv(
...
+                    "Unexpected shape of tensor: {}".format(model.shape),
+                    "Must be scalar or have length equal to total number of faces.",
+                )
+
+            A = sdiag(self.face_areas)
+
+        except NotImplementedError:
+            raise NotImplementedError(
+                f"get_face_inner_product_surface_deriv not implemented for {type(self)}"
+            )
```
jcapriot, line 2766: again put the test for if `face_areas` is implemented seperate fom the rest of the code so it is clear that that is what you are catching the exception from.

discretize/operators/inner_products.py
```diff
@@ -95,6 +98,53 @@ def get_edge_inner_product(  # NOQA D102
...
+        # Isotropic case only
+        if model.size == self.nF:
+            Aprop = self.face_areas * mkvc(model)
+            Av = self.average_edge_to_face
+            M = n_elements * sdiag(Av.T * Aprop)
+        else:
+            raise Exception(
+                "Unexpected shape of tensor: {}".format(model.shape),
+                "Must be scalar or have length equal to total number of faces.",
+            )
```
jcapriot, line 140:
```suggestion
        if model.size != self.nF:
            raise ValueError(
                "Unexpected shape of tensor: {}".format(model.shape),
                "Must be scalar or have length equal to total number of faces.",
            )
        Aprop = self.face_areas * mkvc(model)
        Av = self.average_edge_to_face
        M = n_elements * sdiag(Av.T * Aprop)
```

jcapriot (changes requested):
A few more organizational changes to the code.

discretize/base/base_mesh.py
```diff
@@ -2246,6 +2668,278 @@ def get_edge_inner_product_deriv(
...
+
+            def innerProductDeriv(v=None):
+                if v is None:
+                    warnings.warn(
+                        "Depreciation Warning: TensorMesh.innerProductDeriv."
+                        " You should be supplying a vector. "
+                        "Use: sdiag(u)*dMdprop",
+                        FutureWarning,
+                        stacklevel=2,
+                    )
```
jcapriot, line 2821: I know these are in the `_fastInnerProduct` functions, but you should just not accept v=None, so change v to required argument, and remove all of these warnings.

jcapriot:
Please remove all of the `# NOQA E731`, that error should already be ignored in https://github.com/simpeg/discretize/blob/a230b12105a538d82411f9740fb5760b0698c7f1/.flake8#L40

discretize/base/base_mesh.py
dccowan: done

discretize/base/base_mesh.py
dccowan: done

tests/base/test_tensor_innerproduct.py
dccowan: this error has been added to flake8 exceptions

discretize/base/base_mesh.py
dccowan: Sounds good.

jcapriot (approved):
Looks good!

PR #328, opened 2023-06-16
Use any Python 3 in pre-commit

santisoler:
Unpin Python 3.10 in pre-commit configuration and use any available Python 3 interpreter. This fixes issues when the Python version in the environment is not 3.10.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,12 +3,12 @@ repos:
     rev: 23.1a1
     hooks:
       - id: black
-        language_version: python3.10
+        language_version: python3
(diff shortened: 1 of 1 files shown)


santisoler:
@jcapriot I run into some errors today because the environment I had for discretize wasn't running python 3.10, and `pre-commit` wasn't throwing some errors when trying to create the virtual environments. Since we are pinning the black and flake8 versions, the python version shouldn't be so important.

If you think otherwise, then I suggest pinning a python version in the `environment_test.yml`.

jcapriot:
Yeah, this should be fine. I know I was having issues getting pre-commit to run properly on windows on anything less than 3.10

PR #333, opened 2023-09-06
Add cell_nodes property to TensorMesh

santisoler:
Add a new `cell_nodes` property to `TensorMesh` that mimics the one of `TreeMesh` returning the indices of the nodes that correspond to each cell in the mesh. Add tests for the new feature.

### DIFF
--- a/discretize/tensor_mesh.py
+++ b/discretize/tensor_mesh.py
@@ -1,6 +1,7 @@
 """Module housing the TensorMesh implementation."""
 import itertools
 import numpy as np
+import time
 
 from discretize.base import BaseRectangularMesh, BaseTensorMesh
 from discretize.operators import DiffOperators, InnerProducts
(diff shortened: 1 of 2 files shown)


jcapriot (commented):
The implementation looks good!
The only thing would be that we should likely add a detail to the docstring (both here and in the `TreeMesh` while were at it) about the ordering of the nodes returned by the list, i.e.

```
2 -- 3
|       |
0 -- 1
```
For a 2D mesh.

santisoler:
@jcapriot I just added some notes about the order of the nodes for each cell. Let me know if you like it.

> The implementation looks good!

I just copied and pasted your implementation from the `BasePFSimulation`, so congrats to you haha.

jcapriot (approved):
Thanks, looks good to me know!

PR #334, opened 2023-09-06
Reorganize from "Flat Layout" to "SRC Layout"

omid-b:
## Motivation

I am a Geophysicist at BHP Toronto and wanted to install and test **SimPEG** on my work machine. My work machine is a Windows laptop and there are various limitations when it comes to installation of anything on it. A work-around is to install on **WSL** (~Linux for Windows) and it gives full access to the drive enabling me to install **SimPEG**. When I tried `pip install .` to install **SimPEG** from the source code downloaded from the main repository, it did throw a bunch of errors, most of which related to compilation of the **Cython** and C extensions from discretize. These all happened despite the fact that I have the necessary compilers installed.  Using the same approach on my personal old MacBook and Windows (**WSL**) machines, I faced similar issues. So, the main motivation was to fix my problem, but I realized that I could be more useful, and contribute to fix a couple of more issues for discretize on the way.

## Main changes and tested task list

**Note:** the Azure testing configuration files perhaps should be modified now!

- [X] Changing "flat layout" to "src layout" (see [This Link](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/); the main source codes are not modified, just moved into `src/` )
- [X] Making sure **discretize** gets installed using `pip install .` (tested on WSL)
- [X] Making sure correct python wheels are created using `python setup.py sdist bdist_wheel` (tested on WSL)
- [X] Moving setup/build configurations and metadata information to a `pyproject.toml` file as much as possible for better compatibility and readability (limitation exists for extension files)
- [X] Configure for `abi3` (see [C API Stability](https://docs.python.org/3/c-api/stable.html)); backward/forward compatible wheels for future python releases (next step is perhaps implementing `manylinux`; tested on WSL)


omid-b (commented):
@simpeg/simpeg-developers

jcapriot:
Thanks for putting work into this. We need to get discretize switched over to a `pyproject.toml` file as we consistently get errors due to people missing numpy and cython during a discretize installation from pip in fresh environments.

As you might have noticed, the full tests did not run as it failed the style checks. Check out https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html for hints on the style.

I know I said this on slack, but seeing this here again solidifies my opinion. The change to a `pyproject.toml` file should be a PR all on its own. It will be faster to deal with build issues there (I can already see a few places where there will be issues considering the cpp std=17 flag not being correctly dealt with on windows systems).

Changing discretize to a `src_layout` is a differently scoped thing. Separating these two will allow us to deal with those problems specifically. There's several places in configuration files that assume the current folder layout.

omid-b:
Thanks for taking the time to review this PR.
Following your recommendation, I will close this PR and will break-down the tasks into two upcoming PRs. But before that, isn't the Azure's style check fails due to the relocation of files into `src`  as I did not touch the source codes?

Issue #332, opened 2023-08-30
Install fails with Cython 3.0.2


dwfmarchant:
Looks like the install crashes with Cython 3 that was released last month.  Downgrading to the previous Cython release (v0.29.35) fixes things.

```
(simpeg) ➜  ~ pip list
Package           Version
----------------- -------
Cython            3.0.2
geoana            0.5.0
libdlf            0.2.0
numpy             1.25.2
packaging         23.1
pip               23.2.1
scipy             1.11.2
setuptools        68.0.0
setuptools-scm    7.1.0
tomli             2.0.1
typing_extensions 4.7.1
wheel             0.40.0
(simpeg) ➜  ~ pip install discretize
Collecting discretize
  Using cached discretize-0.9.0.tar.gz (1.1 MB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: numpy>=1.20 in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from discretize) (1.25.2)
Requirement already satisfied: cython>=0.29 in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from discretize) (3.0.2)
Requirement already satisfied: setuptools_scm in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from discretize) (7.1.0)
Requirement already satisfied: scipy>=1.8 in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from discretize) (1.11.2)
Requirement already satisfied: packaging>=20.0 in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from setuptools_scm->discretize) (23.1)
Requirement already satisfied: setuptools in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from setuptools_scm->discretize) (68.0.0)
Requirement already satisfied: typing-extensions in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from setuptools_scm->discretize) (4.7.1)
Requirement already satisfied: tomli>=1.0.0 in ./.virtualenvs/simpeg/lib/python3.10/site-packages (from setuptools_scm->discretize) (2.0.1)
Building wheels for collected packages: discretize
  Building wheel for discretize (setup.py) ... error
  error: subprocess-exited-with-error

  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [67 lines of output]
      Compiling discretize/_extensions/interputils_cython.pyx because it changed.
[...]

PR #337, opened 2023-09-08
Hotfix for #332

omid-b:


omid-b:
@simpeg/simpeg-developers

PR #338, opened 2023-09-11
Pyproject.toml

jcapriot:
This switched the installation method for discretize away from a `setup.py` style to a `pyproject.toml`. The purpose of this file is to build a `discretize` wheel in an isolated python environment, then install that into the local environment. This allows us to cleanly have separate build dependencies, and run-time dependencies.

For `discretize` the build dependencies are:
- numpy
- cython
- setuptools_scm

the run time requirements are simply:
- numpy
- scipy

This has long been a recurring issue that we were mostly solving by telling people to install with conda-forge.

The new build process makes use of a meson build backend (instead of setuptools) to build and install all of the extensions. This is a pretty big change for discretize, however... this is already successfully done with scipy, so I don't think we should have any issues in the long run. I intend on adding more details to the installation steps that we need to take.

PR #341, opened 2023-10-03
Pyproject.toml (2nd method)

omid-b:
Hi Discretize developers team!

This is an alternative to #338 created by @jcapriot . While I appreciate the work in #338, I think this PR follows the "Principle of Simplicity" more closely: "Make it as simple as possible, but not simpler!". 

This PR tries to achieve the same goal as in #338 but not using the `meson.build` approach that is proposed by @jcapriot . Not that I say the `meson.build` approach is wrong, No! It is now used in `SciPy`, so it must be good! But I think `meson.build` method is relatively much more complex and unnecessary in the case of `discretize`. So, the main reason that I created this PR was that this method is relatively much simpler and less modifications and addition of new files were required to the original structure of the project in order to achieve the same goal.

It is ready to be reviewed. So feel free to discuss and ask questions please.

#### Summary of changes:

- from `flat-layout` to `src-layout`
- add `pyproject.toml`
- remove all the metadata from `setup.py` as they are now in `pyproject.toml` (it is still required for the extensions)

As a result of these changes, `discretize` now can be installed using both `pip install .` and `pip install -e .` in fresh virtual python environments and their behavior for both types of installations is as expected!


jcapriot:
Thanks for your contribution, however we have decided to go with the `meson-python` route (#338), not only because it completely removes the `setup.py` file to avoid confusion in the future. But also due to the other benefits it imparts (the editable style installs of extension modules, required explicit listing of installed files).

omid-b:
Thanks for your reply. Okay, fair enough.

PR #347, opened 2024-02-20
Replace deprecated fct `product` by `prod`

prisae:
Resolve
> DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead.


santisoler (approved):
Nice catch @prisae! Thanks for the fix! I'm merging this 🚀

PR #360, opened 2024-05-09
Update use of `numpy`'s random number generators.

jcapriot:
Allow the use of numpy's default random number generator in discretize. Still a bit of work to do to make it complete though.


jcapriot:
Believe this is ready to go now. Only thing to discuss is the name of the keyword. I put it as `rng` instead of `seed` to make it clearer that you can pass a random number generator and not just a seed.

The other thing is that I made it so you can't _easily_  set the random number generator for the error messages. I think it's more interesting to keep getting different random messages. :)

santisoler (approved):
Looks good to me! I noticed that this PR also replaces the `pymatsolver` solvers for the scipy ones. I think it's ok if you want to leave those changes in this PR as long as we add it to the list of changes in the description and in the commit message.

PR #364, opened 2024-07-29
Fix slicer re #363

### LINKED ISSUE #363: 3D Slicer Bug
Confusing stuff.

In here, #261, we fixed this:
```
import discretize
import numpy as np
mesh = discretize.TensorMesh(([1, 1], [1, 1], [1, 1]), (0, 0, 0))
mesh.plot_3d_slicer(np.ones(mesh.n_cells))
```

However, the following still gives a wrong result:
```
import discretize
import numpy as np
mesh = discretize.TensorMesh(([1, 1], [1, 1], [1, 1]), (0, 0, 0))
mesh.plot_3d_slicer(np.zeros(mesh.n_edges_z), v_type='Ez')
```

Figure_1

prisae:
This is really an edge case, not high priority at all (more of cosmetic nature than anything else). Closes #363.


jcapriot:
Hey @prisae, the logic for this choosing of `vmin` and `vmax` was getting a bit convoluted to me. So, I just changed it myself here :). This mimics what matplotlib does internally (Creating a `Normalize()` object if there is no `norm`, and then calling `autoscale_None` on the `norm`. This should make it so all three plots have a consistent `norm` and catch all these odd edge cases.

prisae:
That looks much better @jcapriot, great! I tested it and it works on my side. Not sure why the documentation test is failing... :thinking:

jcapriot:
Ok, I’ll get the doc test fixed and then merge it.

jcapriot:
Okay, a few things. 
* I moved the `plt.show` out of the `plot_3d_slicer` call (Made you unable to interact with the figure before showing it when running from a terminal, or running it as a script).
* Did need to filter out the nans when using autoscale.
* Needed to make sure it was a 1D array going in to it (for some norms... odd that this behavior is not consistent on matplotlib's end).

prisae:
Thanks Joe! I will check tomorrow at work with all my examples, but it looks logic, might just need some adjustments in the examples regarding the `plt.show()`. Might be worth to make a note in that regard in the release notes.

PR #365, opened 2024-08-27
Replace `build` for `python-build` in environment_test.yml

santisoler:
The `build` package in `conda-forge` is outdated, `python-build` should be used instead.


jcapriot:
Outdated now

PR #367, opened 2024-08-30
Add `TreeCell.bounds` and `TreeMesh.cell_bounds` methods

santisoler:
**Work in progress**

Add a new `TreeCell.bounds` method that return the bounds of a cell in
a `TreeMesh`.

- Start drafting `cell_bounds` method


jcapriot:
Let's add the relevant information to the underlying c++ Cell class, so you can access it easier in the `TreeMesh` as well.

santisoler:
> Let's add the relevant information to the underlying c++ Cell class, so you can access it easier in the `TreeMesh` as well.

I agree. I would appreciate if you could help a little bit with that, or just point me in the right direction on how is the best way to do so.

jcapriot:
Branching from #356 would be easiest, as I had put a little bit of the functionality in there already. It lets you just grab the minimum and maximum node for the c_cell pretty easily.

jcapriot:
@santisoler this should be good now

santisoler (commented):
Looks good to me. I just pushed one extra test to make sure that the bounds are returned in the right order. Feel free to merge.

jcapriot (approved):
LGTM, let’s merge

PR #371, opened 2024-09-03
Add version switcher to discretize docs

santisoler:
Configure PyData Sphinx Theme's version switcher and add a new `docs/_static/versions.json` file where versions and their respective urls are listed.


santisoler:
@jcapriot let me know what do you think. Feel free to modify the `versions.json` directly according to the old versions we want to list in the switcher.

jcapriot:
Let's keep it referenced by the actual full name of the tag, so including the "v".
I also adjusted the `conf.py` a bit so as to determine the version information once a bit further up

jcapriot:
Leave the "v" in the "version" section of the switcher json, (you can remove it in the name).
It'll make it easier to point to the correct branch in the deploy step.

docs/_static/versions.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "version": "main",
+        "url": "https://discretize.simpeg.xyz/en/main/"
+    },
+    {
+        "name": "0.10.0 (latest)",
+        "version": "v0.10.0",
```
santisoler, line 8: Actually, the `version` field should not have the leading `v` because these versions should be parsable by the compare-versions node module (https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html#version-warning-banners).

This way, the warning version banners can behave properly. If we have the `v`, any previous version is recognized as a development version, so the banner shows the wrong text.

We can still use the leading `v` in the url though.

docs/conf.py
```diff
@@ -94,8 +95,16 @@
...
 # The full version, including alpha/beta/rc tags.
 release = version("discretize")
+discretize_version = parse(release)
+
+# The short X.Y version.
+version = discretize_version.public
+if discretize_version.is_devrelease:
+    branch = "main"
+else:
+    branch = f"v{version}"
```
santisoler, line 105: We should also remove the leading `v` from here for the version to be parsable.
```suggestion
    branch = f"{version}"
```

santisoler (commented):
Sorry for stepping over your toes in that commit (I didn't see you change). I left two comments explaining why we actually need to remove the leading `v`.

docs/_static/versions.json
jcapriot: `compare_versions` ignores any leading `v`
https://www.npmjs.com/package/compare-versions

santisoler: I actually experienced a misbehaviour in the version banner in simpeg before I changed the versions to not have a leading `v`. But yes, you're right that they seem to support it...

Ignore this then and we'll see if it works well.

jcapriot: It looks like the pydata sphinx theme uses the leading `v` in their's for matching, https://github.com/pydata/pydata-sphinx-theme/blob/main/docs/_static/switcher.json

jcapriot (approved):
Let's do it!

docs/_static/versions.json
santisoler: Then maybe there was something else that was failing back then... Well, I'm glad we don't have to fix it.

PR #372, opened 2024-09-03
Deploy docs to a new folder named after their tagged version

santisoler:
- Make deployment stage to depend on DocBuild
- Store discretize version in a variable
- Update `latest` symlink

TODO:

* [x] Replace the `latest` folder by a symlink in `simpeg/discretize-docs` repo before merging this PR.
* [x] Remove the "experimental" stage I created to check if the version variable can be read in another stage.


jcapriot:
Ok, this should work now. I put in three base pipeline level variables that all stages have access to.
* BRANCH_NAME
* IS_TAG
* IS_MAIN
It creates a folder based on BRANCH_NAME, which is `main` for `main`, and the name of the tag for release versions (e.g. `v0.10.0`).

jcapriot:
Let's merge and see what breaks!

PR #379, opened 2024-10-18
Numpy2.0 updates

jcapriot:
Add some more updates for numpy 2.0,


jcapriot:
Updates look good, the remaining numpy2.0 deprecation warnings are internal to the `omf` package (which is optional and out of our control).

PR #384, opened 2024-10-30
Warn for non-repeatable random tests in a testing environment

jcapriot:
When someone runs a test that would randomly initialize a random number generator from `pytest` or `nosetest`, issue a warning recommending setting the seed for repeatable tests.

This checks by using stack inspection looking for either `_pytest.` or `nose.`

### DIFF
--- a/discretize/tests.py
+++ b/discretize/tests.py
@@ -25,7 +25,6 @@ Functions
   assert_isadjoint
 """  # NOQA D205
 
-import os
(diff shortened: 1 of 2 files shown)


jcapriot:
@santisoler Thought this would be a good thing to do. It will only issue a warning when ran in the testing environment.

santisoler:
I like it! Nicely done.

PR #386, opened 2024-11-20
Set no diagonal balance when reading UBC tree meshes

santisoler:
Set `diagonal_balance=False` when creating a `TreeMesh` after reading a UBC file in `read_UBC`. This prevents raising the future warning related to the default value of `diagonal_balance`. Add test that checks that no warning is being raised by that method.


jcapriot (approved):
Looks good, I believe this catches all the places where TreeMesh's are created internally.

PR #407, opened 2025-10-07
V0.12.0 staging

jcapriot:
Adds release notes and updates version switcher for 0.12.0 release

### DIFF
--- a/docs/release/0.12.0-notes.rst
+++ b/docs/release/0.12.0-notes.rst
@@ -6,7 +6,7 @@
 ``discretize`` 0.12.0 Release Notes
 ===================================
 
-October 7, 2025
+October 8, 2025
 
 This minor release contains many bugfixes and updates related to new package builds.
 
--- a/docs/release/0.12.0-notes.rst
+++ b/docs/release/0.12.0-notes.rst
@@ -1,0 +1,44 @@
+.. currentmodule:: discretize
+
+.. _0.12.0_notes:
+
+===================================
+``discretize`` 0.12.0 Release Notes
+===================================
+
+October 7, 2025
+
+This minor release contains many bugfixes and updates related to new package builds.
+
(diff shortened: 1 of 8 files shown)


docs/release/0.12.0-notes.rst
```diff
@@ -0,0 +1,44 @@
+.. currentmodule:: discretize
+
+.. _0.12.0_notes:
+
+===================================
+``discretize`` 0.12.0 Release Notes
+===================================
+
+October 7, 2025
```
santisoler:
```suggestion
October 8, 2025
```

pyproject.toml
```diff
@@ -89,10 +89,10 @@ style = [
     "flake8-pyproject==1.2.3",
 ]
 build = [
-    "meson-python>=0.14.0",
+    "meson-python>=0.15.0",
     "meson",
     "ninja",
-    "numpy>=1.22.4",
+    "numpy>=2.0.0rc1",
```
santisoler, line 95: Should we mention that we are requiring Numpy >= 2.0 since this release in the release notes?

santisoler, line 95: And just being curious, why choosing 2.0.0rc1 as the minimum supported version and not 2.0.0 directly? Why the release candidate?

santisoler (approved):
Looks good, @jcapriot! I left only a few comments below. Should be ready to go. Thanks for taking care of this!

pyproject.toml
jcapriot: These are specifically `build` requirements, not `runtime` requirements. And this was the first to support the necessary bits for building (this version is ABI compatible with 1.22+ btw, which is why the `runtime` requirement is the same).

santisoler: Great. Marking this as resolved then!

PR #408, opened 2025-10-08
Use ruff as linter and formatter

jcapriot:
Attempts at using ruff to lint and format code instead of flake8 + black


santisoler:
💯 ! I would recommend checking out the Ruff config template that Scientific Python created: https://learn.scientific-python.org/development/guides/style/#ruff

It includes many additional rules that aren't checked by default. Some are too strict, but I found many of them quite useful.

jcapriot:
Sounds good, I’ll add them in and check it out.

PR #412, opened 2026-02-20
Hierarchical inner products for unstructured meshes

dccowan:
This PR is meant primarily to add the hierarchical inner product matrix functionality for 3D simplex meshes. However, the functionality has also been extended for 2D meshes.


discretize/unstructured_mesh.py
```diff
@@ -519,7 +519,64 @@ def get_edge_inner_product(  # NOQA D102
...
+                    Mu = sp.diags(model * face_areas)
+                else:
+                    raise ValueError(
+                        "Unrecognized size of model vector.",
+                        "Must be scalar or have length equal to total number of faces."
+                    )
+                    
+            return Mu        
+            
+        else:
```
jcapriot, line 546:
```suggestion
        if self.dim == 2:
            return super().get_face_inner_product_surface(model=model, invert_model=invert_model, invert_matrix=invert_matrix)
        else:
            if invert_matrix:
                raise NotImplementedError(
                    "The inverse of the inner product matrix with a tetrahedral mesh is not supported."
                )
        
            # Edge inner product surface projection matrices
            n_faces = self.n_faces
            face_areas = self.face_areas
            dim = self.dim
```

jcapriot: removes the code duplication from the default implementation of `get_face_inner_product_surface` in `BaseMesh`

discretize/unstructured_mesh.py
```diff
@@ -519,7 +519,64 @@ def get_edge_inner_product(  # NOQA D102
...
+            Ps = self._get_edge_surf_int_proj_mats(with_area=False)
+        
+            if model is None:
+                Mu = sp.diags(np.tile(face_areas, dim))
```
jcapriot: Just a note that using "dim" here is correct, but a little bit misleading. It's correct because in 3D there are 3 edges per face, and `dim==3`, but it might be more clear to just hard-code a `3` here and add a comment like `3 edges per face`


```suggestion
                Mu = sp.diags(np.tile(face_areas, 3))  # 3 edges per face
```

discretize/unstructured_mesh.py
```diff
@@ -519,7 +519,64 @@ def get_edge_inner_product(  # NOQA D102
...
+            else:
+                if invert_model:
+                    model = 1.0 / model
+                
+                if (model.size == 1) | (model.size == n_faces):
+                    Mu = sp.diags(np.tile(model * face_areas, dim))
```
jcapriot: Same as line 562


```suggestion
                    Mu = sp.diags(np.tile(model * face_areas, 3))  # 3 edges per face
```

tests/simplex/test_inner_products.py
```diff
@@ -412,6 +708,87 @@ def test_EdgeIP_3D_tensor(self):
...
+    def test_EdgeIP_3D_float(self):
+        self.assertTrue(self.doTestEdge([10, 4, 5], 0))
+
+    def test_EdgeIP_2D_isotropic(self):
+        self.assertTrue(self.doTestEdge([10, 4], 1))
+
+    def test_EdgeIP_3D_isotropic(self):
+        self.assertTrue(self.doTestEdge([10, 4, 5], 1))
+
+class TestFacePropertiesInnerProductsDerivs(unittest.TestCase):
```
jcapriot: This class overwrites the test class above it, so that the ones above are never ran.

discretize/curvilinear_mesh.py
jcapriot, line 747: I can't find that this function is actually used by the `CurvlinearMesh` class in a `get_edge_inner_product_surface` function for this class, I believe it just call's the one from `InnerProduct` here: https://github.com/simpeg/discretize/blob/c7763885314e2dd9b7ec28a2b87f31bf81f7a3f1/discretize/operators/inner_products.py#L328

We should also add a corresponding implementation for its `get_edge_inner_product_surface`  at some point. At the very least we should make that function fail in general on the `CurvilinearMesh`

jcapriot (changes requested):
A few changes

jcapriot:
@dccowan I've done a pass through adding the functionality to the `CurvilinearMesh`. Is there anything else that needs to be added?

dccowan:
I can't think of anything else. Thanks for adding the functionality for curvilinear meshes.

PR #416, opened 2026-07-07
Fix bug in mesh refinement with multiple points

### LINKED ISSUE #415: Bug: Unexpected behaviour on `TreeMesh.refine_points`
# The issue

When trying to refine a `TreeMesh` using the `refine_points` method I noticed an unexpected behaviour if the default value of `padding_cells_by_level` is used.

For example, let's consider this simple `TreeMesh` and two points around which I'll refine the mesh: a south-west and a north-west one. In the first snippet I pass the two points to the method and I explicitly state the mesh to refine at the maximum level:

```python
import numpy as np
import discretize


hx = [(1.0, 32)]
mesh = discretize.TreeMesh(h=[hx, hx, hx], origin="CCC", diagonal_balance=True)

# Define two refinement points
points = np.array(
    [
        [-6.0, -6.0, 0.0],  # SW
        [ 6.0,  6.0, 0.0],  # NE
    ]
)

# Refine the mesh using TreeMesh.refine_points
mesh.refine_points(
    points,
    level=-1,
)

mesh.plot_slice(mesh.cell_volumes, normal="z", grid=True, slice_loc=0)
plt.gca().set_aspect("equal")
plt.show()
```

Image

I was expecting the final mesh to be refined around the two points, but apparently the method only refined it around the first point in the collection.

If I pass a positive value to `padding_cells_by_level`, I get the expected behaviour:

```python
import numpy as np
import discretize


hx = [(1.0, 32)]
mesh = discretize.TreeMesh(h=[hx, hx, hx], origin="CCC", diagonal_balance=True)

# Define two refinement points
points = np.array(
    [
        [-6.0, -6.0, 0.0],  # SW
        [ 6.0,  6.0, 0.0],  # NE
    ]
)

# Refine the mesh using TreeMesh.refine_points
mesh.refine_points(
    points,
    level=-1,
    padding_cells_by_level=1,
)

mesh.plot_slice(mesh.cell_volumes, normal="z", grid=True, slice_loc=0)
plt.gca().set_aspect("equal")
plt.show()
```

Image

# Thoughts

I'll try to figure out the source of this issue and submit a patch for it. But, this raised me some questions regarding what's the meaning of having `padding_cells_by_level` equal to `None`. Ultimately, that `None` value will be overwritten by zero. So, what if we only accept numeric values for that argument instead: setting 1 as default if that's the value that makes more sense, or zero with a proper patch to avoid the mentioned behaviour.

What do you think?

santisoler:
Patch `insert_cells` to consider multiple points: fix how the total number of iterations are defined in `insert_cells` to extend the support to multiple points even if a single int is passed to `levels`. Update docs of the method.

Closes #415

### DIFF
--- a/discretize/_extensions/tree_ext.pyx
+++ b/discretize/_extensions/tree_ext.pyx
@@ -1195,6 +1195,20 @@ cdef class _TreeMesh:
         levels = np.require(np.atleast_1d(levels), dtype=np.int32, requirements='C')
         cdef int_t n_points = _check_first_dim_broadcast(points=points, levels=levels)
 
+        if levels.ndim !=1:
+            msg = (
+                "Invalid levels argument with '{levels.ndim}' dimensions. "
+                "It must be a single integer or a 1D array."
+            )
+            raise ValueError(msg)
+        if levels.size > points.shape[0]:
+            msg = (
+                "Invalid levels argument with '{levels.size}' elements. "
+                "It must be a single integer or an arry with the same amount of "
+                "elements as points ('{points.shape[0]}')."
+            )
+            raise ValueError(msg)
+
         cdef double[:, :] cs = points
         cdef int[:] ls = levels
 
@@ -1204,7 +1218,7 @@ cdef class _TreeMesh:
             diagonal_balance = self._diagonal_balance
         cdef bool diag_balance = diagonal_balance
 
-        cdef int_t n_iters = max(ls.shape[0], cs.shape[0])
+        cdef int_t n_iters = max(ls.size, cs.shape[0])
         cdef int_t p_step = cs.shape[0] > 1
         cdef int_t l_step = ls.shape[0] > 1
         cdef int_t i_p=0, i_l=0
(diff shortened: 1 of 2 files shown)


discretize/_extensions/tree_ext.pyx
```diff
@@ -1200,11 +1218,12 @@ cdef class _TreeMesh:
             diagonal_balance = self._diagonal_balance
         cdef bool diag_balance = diagonal_balance
 
+        cdef int_t n_iters = max(ls.size, cs.shape[0])
```
jcapriot: I would just repeat a single integer up to the number of points. Otherwise this would allow you to pass more levels than points and cause a memory error.

santisoler: Thanks @jcapriot! Yes, I agree. I polished the code a bit.

The `_check_first_dim_broadcast(points=points, levels=levels)` will ensure that there cannot be more levels than points (unless a single point is passed). I added an extra error message when a single point and multiple levels are passed, just to let the user know that there might be something off in their inputs.

